PR #2, opened 2018-07-30
Fixing problem with not overriding process variables from .env file

ShelbyZ:
It looks like we are not overriding process.env variables after parsing.

### DIFF
--- a/src/cli/config.js
+++ b/src/cli/config.js
@@ -44,7 +44,7 @@ function getEnvironment(cli, baseDir) {
       const envContent = readFileSync(fullEnvPath, 'utf8');
       const envValues = dotenv.parse(envContent);
       for (const [key, val] of Object.entries(envValues)) {
-        env[key] = val;
+        process.env[key] = val;
       }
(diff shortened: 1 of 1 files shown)


dkundel:
Hey @ShelbyZ,

This is sort of on purpose. I tried to sandbox the environment variables to avoid accidentally rely on env variables that might be set in your local dev environment without knowing it. 

Any reason why you are using `process.env` instead of `context`?

ShelbyZ:
I think this might have been related to my original thought that only values in process.env were appearing accessible from context.  I think it boiled down to a run not reading the configuration file thus not seeing the variables on context.

ShelbyZ:
So the reason I pushed all configuration values into process.env is that the code for is reading directly from process.env for account_sid/auth_token:

```JavaScript
function getTwilioClient() {
  return twilio(process.env.ACCOUNT_SID, process.env.AUTH_TOKEN);
}
```

dkundel:
Good point! I'll fix that in a different way though :)

PR #3, opened 2018-09-13
DOMAIN_NAME in Twilio Functions does not include https://

dbbidclips:
Create the following in a twilio function to see what I mean:

	callback(null, context.DOMAIN_NAME);
};


dkundel:
Thanks for the contribution!

PR #5, opened 2018-09-24
getAssets()['asset'].path gives path instead of filename

dbbidclips:
I'm not sure if this is the best way to fix that issue, but it seemed to do the job for me.

### DIFF
--- a/src/internal/runtime.js
+++ b/src/internal/runtime.js
@@ -27,7 +27,7 @@ function getAssets() {
   const assets = {};
   for (const path of files) {
     const filename = basename(path);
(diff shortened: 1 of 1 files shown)


dkundel:
Hey @dbbidclips,

Thanks for your contribution.

This one has the same issue as in @ShelbyZ's contribution in #6. I'm happy to merge either of you depending on who can add that additional fix :) 

Cheers,
Dominik

dkundel:
Thank you for your contribution, I went ahead and merged @ShelbyZ's fix to push out an update :)

PR #6, opened 2018-10-03
Fn fix

ShelbyZ:
Fixing function path to include absolute similar to issue #5 , but fixing will allow Functions to call other functions.

### DIFF
--- a/src/internal/runtime.js
+++ b/src/internal/runtime.js
@@ -1,5 +1,5 @@
 const twilio = require('twilio');
-const { basename } = require('path');
+const { basename, join } = require('path');
 const { readdirSync } = require('fs');
 const debug = require('debug')('twilio-run:runtime');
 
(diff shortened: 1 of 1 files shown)


dkundel:
Hi @ShelbyZ,

Thanks for fixing this. Can you address my one comment though so that we can make sure it works in Windows just as well? 🙂

Cheers,
Dominik

ShelbyZ:
I'm not sure I was able to find the comment regarding working in Windows.  I'm currently running on a Window machine without issue.

dkundel:
My bad. Forgot to publish the review 🤦‍♂️ Basically just switching to `path.join()`. It's interesting that it actually works in Windows, they might be doing some fixes inside Node.js but could be that people will use that path in a different way. So just to be safe.

dkundel:
Thank you so much!

PR #8, opened 2018-10-03
Fix server to support parsing JSON request body

ShelbyZ:
- If not in place any JSON request will result in an empty event object


dkundel:
Thank you for the contribution @ShelbyZ!

I merged your PR and will release an updated version.

Cheers,
Dominik

PR #10, opened 2019-04-08
Merge .env variables into process.env

the-red:
Hi. 

Using twilio-run with `--env` option, I can use `ACCOUNT_SID` and `AUTH_TOKEN` variables in Twilio authorization.
However, I cannot use `.env` variables in my Functions. 

I'd like to use `.env` in my Functions.
How about merging `config.env` into `proess.env`? :smiley:

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -14,6 +14,7 @@ async function run() {
   const config = await getConfigFromCli(cli);
   debug('Determined configuration: %o', config);
   process.title = config.appName;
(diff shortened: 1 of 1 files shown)


the-red:
Oh, I misunderstood how to use environment variables 😱 
`context.VARIABLE` is correct, not `process.env.VARIABLE`.
https://jp.twilio.com/docs/runtime/functions/invocation#environmental-variables

I will close this pull request.

PR #13, opened 2019-05-09
✨ Improve functions error handling

stefanjudis:
This PR improves the error handling of functions. I was in the situation that my function threw an error which leads to a not found route. This PR catches errors of functions and also propagates the stack trace as a response.

Screenshot 2019-05-09 at 13 50 58

### DIFF
--- a/src/integration.test.js
+++ b/src/integration.test.js
@@ -23,8 +23,17 @@ const availableFunctions = readdirSync(TEST_FUNCTIONS_DIR).map(name => {
 });
 
 function responseToSnapshotJson(response) {
-  const { statusCode, type, body, text, headers } = response;
+  let { statusCode, type, body, text, headers } = response;
   delete headers['date'];
+
+  if (text.startsWith('Error')) {
+    // stack traces are different in every environment
+    // let's not snapshot values that rely on it
+    text = `${text.split('\n')[0]} ...`;
+    delete headers['content-length'];
+    delete headers['etag'];
+  }
+
   return {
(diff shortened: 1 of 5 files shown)


src/integration.test.js
```diff
@@ -23,8 +23,13 @@ const availableFunctions = readdirSync(TEST_FUNCTIONS_DIR).map(name => {
 });
 
 function responseToSnapshotJson(response) {
-  const { statusCode, type, body, text, headers } = response;
+  let { statusCode, type, body, text, headers } = response;
   delete headers['date'];
+
+  if (text.startsWith('Error')) {
```
stefanjudis: Cutting the stack trace to not have file paths in there.

src/integration.test.js
```diff
@@ -38,7 +43,6 @@ describe('Function integration tests', () => {
   for (const testFnCode of availableFunctions) {
     test(`${testFnCode.name} should match snapshot`, async () => {
       const response = await request(app).get(testFnCode.url);
-      expect(response.statusCode).toBe(200);
       const result = responseToSnapshotJson(response);
```
stefanjudis: It's not only 200 but 200 and 500 now. :)

ShelbyZ:
This is definitely an improvement over needing to wrap all developed functions with try/catch to see what error is happening locally.

dkundel:
Thanks for the contribution! I had this fix already in the next branch for v2 but it's great that you ported it to v1

PR #14, opened 2019-05-10
Fixing issue with getSync in runtime

ShelbyZ:
- needed default serviceName not object destructuring

### DIFF
--- a/src/internal/runtime.js
+++ b/src/internal/runtime.js
@@ -51,9 +51,10 @@ function getFunctions() {
 }
 
 function create({ env }) {
-  function getSync(serviceName = 'default') {
+  function getSync(syncService) {
+    const syncInfo = syncService || { serviceName: 'default' };
     const client = twilio(env.ACCOUNT_SID, env.AUTH_TOKEN);
-    return client.sync.services(serviceName);
+    return client.sync.services(syncInfo.serviceName);
   }
 
   return { getSync, getAssets, getFunctions };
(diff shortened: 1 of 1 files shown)


dkundel:
Hi @ShelbyZ sorry for the delay here but I think at least according to the docs you are supposed to pass in `getSync({ serviceName: 'something' })` and therefore it should be destructured here.

https://www.twilio.com/docs/runtime/client?code-sample=code-get-an-existing-sync-document-12&code-language=Node.js&code-sdk-version=default

ShelbyZ:
It looks like that line of code is trying to provide a default value for serviceName.  The call does not set the value of serviceName to 'default' so later ```return client.sync.services(serviceName);``` will have undefined for the value of serviceName.

dkundel:
Oh yeah my bad. So the code should be

```
  function getSync(syncService) {
    syncInfo = syncInfo || { serviceName: 'default' };
    const client = twilio(env.ACCOUNT_SID, env.AUTH_TOKEN);
    return client.sync.services(serviceName);
  }
```

ShelbyZ:
```function getSync(syncService) {
    syncInfo = syncService || { serviceName: 'default' };
    const client = twilio(env.ACCOUNT_SID, env.AUTH_TOKEN);
    return client.sync.services(syncInfo.serviceName);
  }```

The call to ```client.sync.services(syncInfo.serviceName);``` - is either 'default' or the sync service sid

dkundel:
Yeah that should be the behavior

dkundel:
Thank you so much for the fix! I'll publish a new version of 1.x later 😊

PR #56, opened 2019-07-30
refactor(logging): replace console with central logger (#50) 

### LINKED ISSUE #50: Replace all `console.` calls with a centralized logger
This way we can implement log levels for `info`, `warn` and `error` better and centralize the output style. Debug logs should remain separate in the files and keep relying on the `debug` module. Instead if the log level is set to debug, the appropriate environment flags (`DEBUG=twilio*`) should be enabled.

dkundel:
In an attempt to make this more compatible with the Twilio CLI in the future, this commit refactors the code the following way:
- we can swap in other loggers
- introduces configurable log-levels via command
- introduces a `writePlainOutput` that should be the only function to be used to write output to `stdout`.
- improves the logical separation between command-output (result) and info/warn/error messages

From now on no `console.*` should be used outside of `utils/output.ts`. 

Fix #50 

**Contributing to Twilio**


dkundel:
Hey @philnash this shouldn't break any of your code since the default log level is the same one as before, but I just wanted to make you aware of it. It would be great if you could have a look over it before I merge it in.

src/utils/output.ts
```diff
@@ -0,0 +1,3 @@
+export function writePlainOutput(...args: any[]) {
```
philnash: I'm not sure this is the best name to explain that this function is the one to use to show output to a user. It also doesn't need to be used for "plain" output, as there are uses in this PR where `chalk` coloured text are sent to output.

Would it fit better as part of the logger class too? As discussed, all the log levels (debug/info/etc) pipe to stderr, and this function is the only one sending to stdout, but for other developers reaching to use the logger or the outputter, this might be a bit lost on its own. Might we consider `logger.output(...args)` as a method on the `logger` object that specifically will output things, whilst everything else is logged?

Or is that just complicating things more?

src/templating/actions.ts
```diff
@@ -10,9 +11,9 @@ export async function downloadTemplate(
   const files = await getTemplateFiles(templateName);
   try {
     await writeFiles(files, targetDirectory, bundleName);
-    console.log(chalk`{green SUCCESS} Created new bundle ${bundleName}`);
+    logger.info(chalk`{green SUCCESS} Created new bundle ${bundleName}`);
```
philnash: Should this use `writePlainOutput` instead of logging as it is a specific output to the user?

src/templating/filesystem.ts
```diff
@@ -162,7 +162,7 @@ export async function writeFiles(
 
   const newKeys = context.env.newEnvironmentVariableKeys;
   if (newKeys.length > 0) {
-    console.log(
+    logger.info(
```
philnash: Should this use `writePlainOutput` instead of logging as it is a message to the user?

dkundel: No because in that case you want to write to `stderr` and keep `stdout` empty. Here's a good example. https://www.jstorimer.com/blogs/workingwithcode/7766119-when-to-use-stderr-instead-of-stdout

src/utils/output.ts
dkundel: I think this is complicating things more and the reason why I'm keeping the two separate is because that's how the Twilio CLI does it as well and if I'm playing around with this, I might as well create parity. 
The idea is that output is really only used for the "result". So if a command fails, the result should be empty for example.
The reason I introduced this to be more clear when to use output and when to log something. But I agree the naming might not be great. Maybe it's just `writeOutput`.

src/templating/filesystem.ts
philnash: Ok, good article! :+1: I'm convinced.

src/utils/output.ts
philnash: `writeOutput` definitely works better for me than `writePlainOutput` 👍

PR #75, opened 2019-08-15
fix(activate): environment was 'required' not 'requiresArg'

ShelbyZ:
activate command had incorrect 'required' property which prevented treatment as a required argument when converted from yargs to oclif.

**Contributing to Twilio**


dkundel:
Thank you for jumping on the fix for this!

PR #80, opened 2019-09-22
fix(server): object Runtime not available before function run

ShelbyZ:
It is possible to define a function that uses the Runtime prior to handler execution:

```JavaScript
const hello = require(Runtime.getFunctions()['hello'].path);

    hello.handler(context, event, () => {
        console.log('ran');
        callback(null, {});
    });
};
```

The above code will work when deployed to Twilio, but not when hosted via twilio-run.

**Contributing to Twilio**


dkundel:
Thanks for the contribution! Great job

PR #81, opened 2019-09-22
feat: add estimated execution time to function run

ShelbyZ:
- use process.hrtime() to track start/stop (supports nodeJS 8.x)
- debug log estimated execution time in milliseconds

Not sure if requiring debug() is the correct way to report this information.  We could also consider displaying a warning/error message if the estimated duration execeeds the defined limit.

Another approach to capturing timing would be to use perf_hooks to setup an observer and add measurements.  The observer could process the measurements (logging) and disconnect.

**Contributing to Twilio**

### DIFF
--- a/src/runtime/route.ts
+++ b/src/runtime/route.ts
@@ -147,17 +147,17 @@ export function functionToRoute(
     ) {
       run_timings.end = process.hrtime();
       debug('Function execution %s finished', req.path);
-      if (err) {
-        handleError(err, req, res, functionFilePath);
-        return;
-      }
-      handleSuccess(responseObject, res);
       debug(
         `(Estimated) Total Execution Time: ${(run_timings.end[0] * 1e9 +
           run_timings.end[1] -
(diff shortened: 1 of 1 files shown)


ShelbyZ:
Hold off on this for now, I may convert this to use `process.hrtime()` since node 8.10 does not have `process.hrtime.bigint()`.

src/runtime/route.ts
```diff
@@ -133,21 +133,36 @@ export function functionToRoute(
...
       responseObject?
     ) {
+      run_timings.end = process.hrtime();
       debug('Function execution %s finished', req.path);
       if (err) {
         handleError(err, req, res, functionFilePath);
         return;
       }
       handleSuccess(responseObject, res);
+      debug(
```
dkundel: Should we have this on line 150 before the error handler? I feel like this is useful even if an error happens as the error might be intentional

ShelbyZ: Yes, I did overlook the return statement on 152, oops.

PR #86, opened 2019-10-10
feat(templates): report GitHub api error

### LINKED ISSUE #68: Better error message for errors with the GitHub API
I just encountered an error with the `new` command where it would throw a random unhandled rejection error saying this:

```
DEBUG=twilio-run:new:template-data twilio serverless:new hello --template=blank                                                           ⬢ 8.10.0
  twilio-run:new:template-data Response code 403 (Forbidden) +0ms
(node:66826) UnhandledPromiseRejectionWarning: Error: Invalid template
    at Object.<anonymous> (/Users/dkundel/.twilio-cli/node_modules/twilio-run/dist/templating/data.js:71:19)
    at Generator.throw (<anonymous>)
    at rejected (/Users/dkundel/.twilio-cli/node_modules/twilio-run/dist/templating/data.js:5:65)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
(node:66826) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function with
out a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:66826) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will t
erminate the Node.js process with a non-zero exit code.
```

After some investigation I realized it was because the GitHub API was returning a `403 - Forbidden` response because the IP I was on, was Rate Limited. We should probably handle errors like that better and just forward the error message in the body by parsing the response body.

Here's an example payload from GitHub:

```json
{
  "message": "API rate limit exceeded for 19.253.99.123. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)",
  "documentation_url": "https://developer.github.com/v3/#rate-limiting"
}
```

anchepiece:
This change surfaces the descriptive message returned when an error is encountered using GitHub API to import templates. 

Tested with status 403 rate limited exceeded and 404 page not found. Suggestions welcome.

Fixes #68 

**Contributing to Twilio**

- [X] I acknowledge that all my contributions will be made under the project's license.

### DIFF
--- a/src/templating/actions.ts
+++ b/src/templating/actions.ts
@@ -8,8 +8,8 @@ export async function downloadTemplate(
   namespace: string,
   targetDirectory: string
 ): Promise<void> {
-  const files = await getTemplateFiles(templateName);
   try {
+    const files = await getTemplateFiles(templateName);
     await writeFiles(files, targetDirectory, namespace);
     logger.info(
       chalk`{green SUCCESS} Downloaded new template into the "${namespace}" subdirectories`
--- a/src/templating/data.ts
+++ b/src/templating/data.ts
@@ -47,6 +47,11 @@ type RawContentsPayload = {
   };
 }[];
 
+type GitHubError = {
+  message: string;
+  documentation_url?: string;
+};
+
 async function getFiles(
   templateId: string,
   directory: string
@@ -97,6 +102,14 @@ export async function getTemplateFiles(
     return files;
   } catch (err) {
     debug(err.message);
+
+    if (err.response) {
+      const bodyMessage = err.response.body as GitHubError;
+      throw new Error(
+        bodyMessage ? `${err.message}\n${bodyMessage.message}` : err.message
+      );
+    }
+
     throw new Error('Invalid template');
   }
 }


dkundel:
Adding @philnash because I'd love to have his opinion on this. This is a good start for unknown errors from the API. I'd love to handle one case specifically:
If the rate limit error is thrown the error message they get is not the most useful one for people. I'd be in favor of catching that one specifically and replying with something like this:

```
We are sorry but we failed fetching the requested template from GitHub because your IP address has been rate limited. Please try the following to resolve the issue:
- Change your WiFi or make sure you are not connected to a VPN that might cause the rate limiting

If the issue persists you can try one of the two options:
- Get a GitHub developer token following https://help.github.com/en/articles/creating-a-personal-access-token-for-the-command-line (no permissions needed) and add it as `TWILIO_SERVERLESS_GITHUB_TOKEN` to your environment variables
- Wait for a few minutes up to an hour and try again.
```

For this we'd have to make sure to pass `TWILIO_SERVERLESS_GITHUB_TOKEN` to the GitHub API if it's available. @philnash what do you think?

philnash:
This sounds like a great idea. We can definitely support optionally sending a GitHub token if it's available. Does that mean we could merge this and open a new issue for the special handling of the rate limit issue and token?

Issue #53, opened 2019-07-29
Watch mode


dkundel:
Local dev mode should have a `watch` flag that automatically adds newly created files to the files that are being served. It could either just restart the entire server or just update the Routes cache and print a message to the terminal that a new Function or Asset has become available.

dsnair:
Is it okay if I give this a try?

philnash:
Please go ahead @dsnair! I don't believe anyone else is working on this and we would love the contribution. Let us know if you'd like any guidance on the project.

imnotjames:
Given there's already a `live` parameter, should it work alongside that?  EG, a filesystem watcher that clears the route caches when a new file is added / removed when `--live` is specified?

As it stands, `--live` seems to disable any sort of caching of functions (and there doesn't seem to be a good way to disable that).

```
  --live                Always serve from the current functions (no caching)
                                                       [boolean] [default: true]
```

If it should be added to a separate `--watch` that's not too bad to do, just seems to be overlapping features.

PR #94, opened 2019-10-21
feat: support loading of new files automatically with `--live`

imnotjames:
Implements a "watcher" on files being added and removed when `--live` is specified in the command line arguments.

**Contributing to Twilio**


imnotjames:
This is a second attempt at #53 now that I understand the problem a bit better :)

src/runtime/server.ts
```diff
@@ -8,6 +8,9 @@ import express, {
 } from 'express';
 import userAgentMiddleware from 'express-useragent';
 import nocache from 'nocache';
+import {printRouteInfo} from '../printers/start';
+import watch from 'node-watch';
```
dkundel: Could you try this library instead? https://www.npmjs.com/package/chokidar? It seems like this is a defacto standard. It's being used by libraries like nodemon, babel and browser-sync

src/runtime/server.ts
```diff
@@ -77,8 +80,44 @@ export async function createServer(
...
+  let routes = await getFunctionsAndAssets(config.baseDir);
+  let routeMap = setRoutes(routes);
+
+  if (config.live) {
+    let seenFiles = new Set();
+
+    watch(config.baseDir, { recursive: true }, async (evt, name) => {
+      let relativeName = path.relative(config.baseDir, name);
+
+      if (!relativeName.match(/^(assets|functions)\//)) {
```
dkundel: If you could check out the library mentioned above, it seems like it supports creating a glob pattern. So we could be even more precise. Something like:
```
chokidar.watch(`${config.baseDir}/{functions/**/*.js,assets/**/*}`)
  .on('add', ...)
  .on('remove', ...)
```

src/runtime/server.ts
imnotjames: That's a much better library to use for this use case - I didn't realize that babel was using that!  I'll definitely make that change.

src/runtime/server.ts
```diff
@@ -77,8 +80,44 @@ export async function createServer(
...
+        // Only reload on assets and functions.
+        // No other files are very important watching.
+        return;
+      }
+
+      if (evt == 'update') {
+        if (name in seenFiles) {
```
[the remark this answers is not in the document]
imnotjames: Odd - I'd investigate it but given that I'm switching this to chokidar we should be fine.

src/runtime/server.ts
```diff
@@ -77,8 +80,44 @@ export async function createServer(
...
+    watch(config.baseDir, { recursive: true }, async (evt, name) => {
+      let relativeName = path.relative(config.baseDir, name);
+
+      if (!relativeName.match(/^(assets|functions)\//)) {
+        // Only reload on assets and functions.
+        // No other files are very important watching.
+        return;
+      }
+
+      if (evt == 'update') {
```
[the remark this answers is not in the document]
imnotjames: No longer relevant by switching to chokidar.

src/runtime/server.ts
imnotjames: I've done this.  I used instead:

```
    const watcher = chokidar.watch(
      [
        path.join(config.baseDir, '/functions/**/*.js'),
        path.join(config.baseDir, '/assets/'),
      ],
...
    )
```

Issue #90, opened 2019-10-17
Allow to set an optional GitHub API token to circumvent rate limiting


dkundel:
In #86 we added a better error message for GitHub API failures. 

However, if the rate limit error is thrown the error message they get is not the most useful one for people. I'd be in favor of catching that one specifically and replying with something like this:

```
We are sorry but we failed fetching the requested template from GitHub because your IP address has been rate limited. Please try the following to resolve the issue:
- Change your WiFi or make sure you are not connected to a VPN that might cause the rate limiting

If the issue persists you can try one of the two options:
- Get a GitHub developer token following https://help.github.com/en/articles/creating-a-personal-access-token-for-the-command-line (no permissions needed) and add it as `TWILIO_SERVERLESS_GITHUB_TOKEN` to your environment variables
- Wait for a few minutes up to an hour and try again.
```

For this we'd have to make sure to pass `TWILIO_SERVERLESS_GITHUB_TOKEN` to the GitHub API if it's available.

badsketch:
Hey, I'd be interested in tackling this issue @dkundel

dkundel:
Yes @badsketch pleae go ahead with this :) Thank you!

badsketch:
@dkundel Does this include adding `--github-token` as a flag option for the `new` command?

On one hand it adds flexibility to how a user uses the cli. On the other hand, it might be out of place since the only time they see it is if they're using `new` extensively to the point of rate limiting. And at that point, they would probably add it to `.env` as a permanent solution instead of flag as a one time use.

I'm still new to contributing to a CLI helper so I don't know all of the use cases. Thoughts? 🙂

badsketch:
Oh, I think I misread what environmental variables meant. Aside from adding the api variable in the `.env` file, `TWILIO_SERVERLESS_GITHUB_TOKEN=xxxx twilio-run new ` would already be a way to specify environment variables via the command line so I guess there wouldn't be a need to add `--github-token` as an option flag right?

dkundel:
Yeah just using the env variable would be totally sufficient. Don't think we need to have a separate flag. Thank you :) and sorry for the delay. I was traveling.

badsketch:
No worries  🙂Thanks for the clarification

PR #96, opened 2019-11-16
feature: #90 allow users to specify a github token

badsketch:
Adds #90 functionality by allowing the user to circumvent rate limiting with an auth token specified in their environment.

First time contributing  to a CLI, so looking for suggestions on improvement. 🙂

**Contributing to Twilio**

### DIFF
--- a/src/templating/data.ts
+++ b/src/templating/data.ts
@@ -1,7 +1,8 @@
 import got from 'got';
 import { getDebugFunction } from '../utils/logger';
 import { stripIndent } from 'common-tags';
-
+import { readLocalEnvFile } from '../config/utils/env';
+import { OutgoingHttpHeaders } from 'http';
 const debug = getDebugFunction('twilio-run:new:template-data');
 
 const TEMPLATES_URL =
@@ -57,8 +58,10 @@ async function getFiles(
   templateId: string,
   directory: string
 ): Promise<TemplateFileInfo[]> {
+  const headers = await buildHeaders();
   const response = await got(CONTENT_BASE_URL + `/${templateId}/${directory}`, {
     json: true,
+    headers,
   });
   const repoContents = response.body as RawContentsPayload;
   return repoContents.map(file => {
@@ -74,8 +77,10 @@ export async function getTemplateFiles(
   templateId: string
 ): Promise<TemplateFileInfo[]> {
   try {
+    const headers = await buildHeaders();
     const response = await got(CONTENT_BASE_URL + `/${templateId}`, {
       json: true,
+      headers,
     });
     const repoContents = response.body as RawContentsPayload;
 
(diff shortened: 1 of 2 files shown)


src/templating/data.ts
```diff
@@ -1,7 +1,8 @@
 import got from 'got';
 import { getDebugFunction } from '../utils/logger';
 import { stripIndent } from 'common-tags';
-
+import { readLocalEnvFile } from '../config/utils/env';
+import { OutgoingHttpHeaders } from 'http';
```
badsketch: got() options required headers of type `OutgoingHttpHeaders`, but I wonder if this was the correct way of handling it or I'm just not using TypeScript correctly?

src/templating/data.ts
```diff
@@ -131,3 +136,18 @@ export async function getTemplateFiles(
     throw new Error('Invalid template');
   }
 }
+
+async function buildHeaders(): Promise<OutgoingHttpHeaders> {
+  let githubToken = '';
+  const { localEnv } = await readLocalEnvFile({ cwd: process.cwd() });
```
badsketch: I noticed typical usage of `readLocalEnvFile` included the cwd and env from flags. Should I have also done the same?

dkundel: I would actually use `process.env.TWILIO_SERVERLESS_GITHUB_TOKEN` instead of reading the local env file for this. Using `readLocalEnvFile` would enable people to place this into `.env` files but that would also cause that variable to be uploaded to Twilio Functions if they ran `twilio-run deploy`. Instead I think this token should live on your dev machine only. So using `process.env` is the right approach here.

src/types/generic.ts
```diff
@@ -3,4 +3,5 @@ import { EnvironmentVariables } from '@twilio-labs/serverless-api';
 export type EnvironmentVariablesWithAuth = EnvironmentVariables & {
   ACCOUNT_SID?: string;
   AUTH_TOKEN?: string;
+  TWILIO_SERVERLESS_GITHUB_TOKEN?: string;
```
dkundel: If you implement my previous comment you can skip this step :)

src/templating/data.ts
```diff
@@ -104,12 +110,44 @@ export async function getTemplateFiles(
     debug(err.message);
 
     if (err.response) {
-      const bodyMessage = err.response.body as GitHubError;
-      throw new Error(
-        bodyMessage ? `${err.message}\n${bodyMessage.message}` : err.message
-      );
+      if (err.response.statusCode === 403) {
+        throw new Error(
+          stripIndent`
```
dkundel: Great description!

src/templating/data.ts
dkundel: Yeah this should be fine!

badsketch:
@dkundel No worries I understand! Thanks for taking the time to offer thoughtful comments. I'll address these issues.  Cheers 🎉

PR #100, opened 2020-01-10
feat(server): expose function path to context

dkundel:
This exposes the current function path as the PATH property on the context. This catches up with the
release of that variable in the Twilio Functions environment as introduced in
twilio.com/changelog/twilio-functions-context-path

I marked this as a breaking change because if someone used `PATH` as an env variable before this will not work anymore. 

**Contributing to Twilio**

### DIFF
--- a/examples/basic/functions/hello.js
+++ b/examples/basic/functions/hello.js
@@ -3,5 +3,6 @@
 exports.handler = function(context, event, callback) {
   let twiml = new Twilio.twiml.MessagingResponse();
   twiml.message('Hello World');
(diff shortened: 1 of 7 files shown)


examples/basic/functions/hello.js
```diff
@@ -3,5 +3,6 @@
 exports.handler = function(context, event, callback) {
   let twiml = new Twilio.twiml.MessagingResponse();
   twiml.message('Hello World');
+  console.log(context);
```
philnash: Did you mean to leave this in?

PR #105, opened 2020-01-15
feat(server): hide internals from stack trace

dkundel:
By default internals from the twilio-run dev server will no-longer be part of the stack trace.
Instead we'll display them as Twilio Dev Server internals. More info in the README

**Contributing to Twilio**


philnash:
This is less code than I thought it was going to be!

Issue #62, opened 2019-08-02
Implement machine readable output format


stefanjudis:
I'm currently writing an automation shell script that buys a number, creates a sync services, and deploys a function.

The twilio CLI provides the command option `-o=json` for several other commands. This option allows to get the command response in a machine readable way. This makes it possible to re-use values write advanced scripts. It would be great to have this option in the Twilio Serverless CLI plugin, too. 

As the CLI is only a wrapper around `twilio-run` i thought that it's best to open the issue here first. :) 

What I'm looking for is something as follows:

```
# in the Twilio CLI
$ twilio serverless.deploy -o=json

# and in twilio-run
$ twilio-run deploy --output=json
```

Is this something you'd be up for?

dkundel:
This is definitely on the roadmap for post SIGNAL. Thanks for filing this.

philnash:
OK, so I've been working on this and I got quite a way into it before getting confused 😄

The commands `list`, `list-templates`, `deploy`, `activate` and `logs` all call off to APIs and have JSON results.

The commands `new` and `start` do not do that and don't really have an obvious return data. `start` isn't really a command that would be used by a machine but `new` is used by `create-twilio-function`.

So far, the work is ongoing on [my json-output branch](https://github.com/twilio-labs/twilio-run/compare/master...philnash:json-output). I've added a `output-format` option to the `BaseCliFlags` but now I realise that the `start` and `new` commands don't necessarily need a JSON output, so perhaps it shouldn't be in `BaseCliFlags`.

So, to my questions:

Should the `start` and `new` commands have the `--output-format` flag as an option at all?

Should those two have another flag/option which silences their output so that other programs can use them without them making a mess over the terminal (e.g. see the test output of `create-twilio-function`)?

Or should I just leave the `start` and `new` commands alone for now?

PR #110, opened 2020-02-11
feat(logs): adds command to access new logging functions

philnash:
Initial attempt to add logs to twilio-run.

Much like [the PR in serverless-api](https://github.com/twilio-labs/serverless-api/pull/23) this doesn't cover all the filters available for the logs endpoint. It is a first attempt and could do with a review.

It adds the logs command.

```bash
npx twilio-run logs --tail
```

The above command will pick up the project config and stream the logs from the functions in the service and environment (dev by default).

**Contributing to Twilio**


src/printers/logs.ts
```diff
@@ -0,0 +1,11 @@
+import { LogList, LogApiResource } from '@twilio-labs/serverless-api';
+import { LogsConfig } from '../config/logs';
+import { logger } from '../utils/logger';
+
+export function printLogs(result: LogApiResource[], config: LogsConfig) {
+  result.forEach(printLog);
+}
+
+export function printLog(log: LogApiResource) {
+  logger.info(`[${log.level}][${log.date_created}]: ${log.message}`);
```
dkundel: Let's future proof this. #62 is calling for the same support as the Twilio CLI in terms of output. Let's add support for an `--output` or `-o` flag that supports at least `json` as another option where each line will be it's own line of JSON. That way you could pipe it into a command to split by line and then run `jq` on it to filter for certain things. 
If you want you could add `tsv` support as well but I think `json` is for now the only thing.

philnash: Ah yeah, I actually meant to do this. Being able to output JSON particularly so that the whole object can be piped to something else to render differently.

philnash: I've added this now.

philnash:
Travis CI seems to be having issues with installing/running Node 10. @dkundel Can you nudge Travis and kick [the build](https://travis-ci.org/github/twilio-labs/twilio-run/jobs/662871518) off again to make it pass please.

src/printers/logs.ts
```diff
@@ -0,0 +1,19 @@
...
+  result: LogApiResource[],
+  config: LogsConfig,
+  outputFormat?: string
+) {
+  result.forEach(log => printLog(log, outputFormat));
+}
+
+export function printLog(log: LogApiResource, outputFormat?: string) {
+  if (outputFormat === 'json') {
+    logger.info(JSON.stringify(log));
```
dkundel: `logger.info` will log to `stderr` not to `stdout`. You should instead use `writeOutput` from the `../utils/output` module.

src/config/logs.ts
```diff
@@ -0,0 +1,77 @@
...
+  environment = flags.environment || environment;
+
+  const { accountSid, authToken } = await getCredentialsFromFlags(
+    flags,
+    externalCliOptions
+  );
+
+  const command = getFullCommand(flags);
+  const serviceSid = checkForValidServiceSid(command, flags.serviceSid);
+  const tail = !!flags.tail;
```
dkundel: What's the logic behind this? Should we throw an error instead if it's not a `boolean`? I feel like this could create accidentally wrong situations if the parser turns something like this `--tail` into `tail: ''` and that would turn into false.

philnash: You're right, this is because I was not relying on the types. This line can be removed.

src/printers/logs.ts
philnash: Thanks! Thought I was being good using `logger.info`. Updated.

philnash:
I believe this to be complete now. My concerns about not showing logs were unfounded.

PR #116, opened 2020-03-13
feat(templates): template README downloaded and saved to readmes dir

### LINKED ISSUE #104: README not copied when using --template
The README from the function template does not get copied across during initialization from a template.

It would be ideal to be able to use README's for functions that require additional setup/instructions.

Definitely open to other ways of providing setup instruction as well.

philnash:
Fixes #104

**Contributing to Twilio**

### DIFF
--- a/src/templating/filesystem.ts
+++ b/src/templating/filesystem.ts
@@ -85,7 +85,8 @@ function hasFilesOfType(files: TemplateFileInfo[], type: string) {
 export async function writeFiles(
   files: TemplateFileInfo[],
   targetDir: string,
-  namespace: string
+  namespace: string,
+  templateName: string
 ): Promise<void> {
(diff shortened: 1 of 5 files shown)


src/templating/filesystem.ts
```diff
@@ -156,6 +167,17 @@ export async function writeFiles(
           title: 'Installing Dependencies',
           task: () => installDependencies(file.content, targetDir),
         };
+      } else if (file.type === 'README.md') {
+        const readmePath = path.join(readmesTargetDir, namespace + '.md');
```
dkundel: Since we added support for writing multiple templates to the same namespace, should this name include the template name?

dkundel: Also how will this play with `create-twilio-function` where the namespace is not existing

philnash: This should all work now!

PR #117, opened 2020-03-14
fix: #111 change pinned node version to 10.x

### LINKED ISSUE #111: Change pinned Node.js version
Right now the version we check for is pinned to `10.17.0`. The cloud environment of Twilio Functions doesn't actually pin to a specific minor version anymore so we should change this to just check for `10.x`.

badsketch:
Currently, we check if the node version is 10.17. Twilio Functions cloud no longer checks minor
versions so we only need to check for 10.x

fix #111

**Contributing to Twilio**

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,8 +12,8 @@ This project contains the CLI aspects of deploying Twilio Serverless as well as 
 
 ## Requirements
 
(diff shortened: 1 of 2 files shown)


badsketch:
Hope I understood the issue correctly. Looking for any feedback 🙂🚀

PR #123, opened 2020-03-30
feat(templates): support nested templates to create nested routes

philnash:
Fixes #122

This adds a `directory` property to `TemplateFileInfo` and then uses that to place the files in the correct nested directories.

**Contributing to Twilio**

> All third-party contributors acknowledge that any contributions they provide will be made under the same open-source license that the open-source project is provided under.

- [x] I acknowledge that all my contributions will be made under the project's license.

PR #125, opened 2020-04-07
fix(templates): adds tests for getFiles and replaces Array.flat

### LINKED ISSUE #124: Re-add support for Node 10
With #123 we broke support for Node 10 because it's missing the `Array.prototype.flat` method that was introduced only in Node 11. We should also investigate why the tests didn't catch this.

philnash:
#123 broke support in Node < 11 due to using `Array#flat` 🤦‍♂️

This adds tests to ensure the `getFiles` function works properly and replaces `Array#flat` with `lodash.flatten`.

Fixes #124.

**Contributing to Twilio**


philnash:
Squashed and released.

PR #127, opened 2020-04-14
feat(start): handle EADDRINUSE by asking for another port number

### LINKED ISSUE #97: Handle EADDRINUSE error
We should consider offering an option to use another port when starting twilio-run with the default or supplied port is in use.

```
events.js:174
      throw er; // Unhandled 'error' event
      ^

Error: listen EADDRINUSE: address already in use :::5000
```

philnash:
If the server errors with `EADDRINUSE` then we ask for a new port number with the default of the existing port number + 1.

Fixes #97.

**Contributing to Twilio**

### DIFF
--- a/src/commands/start.ts
+++ b/src/commands/start.ts
@@ -1,4 +1,5 @@
 import { Argv } from 'yargs';
+import inquirer from 'inquirer';
 import checkNodejsVersion from '../checks/nodejs-version';
 import checkProjectStructure from '../checks/project-structure';
 import { getConfigFromCli, StartCliFlags } from '../config/start';
@@ -11,6 +12,10 @@ import { CliInfo } from './types';
 import { getFullCommand } from './utils';
 
 const debug = getDebugFunction('twilio-run:start');
+
+type ServerError = Error & {
+  code: string;
+};
 
 export async function handler(
   argv: StartCliFlags,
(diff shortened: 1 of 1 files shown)


src/commands/start.ts
```diff
@@ -41,11 +46,44 @@ export async function handler(
...
+        ]);
+        const server = app.listen(
+          answers.newPortNumber,
+          serverStartedSuccessfully
+        );
+        server.on('error', handleServerError);
+      } else {
+        reject(error);
+      }
+    };
```
ShelbyZ, line 83: Should we care about placing a max limit on the number of times to prompt a user for an available port?

src/commands/start.ts
```diff
@@ -41,11 +46,44 @@ export async function handler(
...
       printRouteInfo(config);
       resolve();
-    });
+    };
+    const handleServerError = async (error: ServerError) => {
+      if (error.code === 'EADDRINUSE') {
+        const answers = await inquirer.prompt([
+          {
+            type: 'input',
+            default: config.port + 1,
```
ShelbyZ, line 59: Do we need to valid that +1 will keep the port within 0..65535?

src/commands/start.ts
philnash: Good point, I've added a maximum of 3 retries then exit with an error message. Thanks.

src/commands/start.ts
philnash: Also good point! I've made this a random port number now.

src/commands/start.ts
```diff
@@ -9,9 +10,19 @@ import { getDebugFunction, setLogLevelByName } from '../utils/logger';
...
 
 const debug = getDebugFunction('twilio-run:start');
 
+type ServerError = Error & {
+  code: string;
+};
+
+function randomPort() {
+  // Returns a random port number higher than 1024 and lower than 65536.
+  return Math.floor(Math.random() * (65535 - 1025) + 1025);
```
dkundel, line 23: Should we start by trying out 3001? I know other tools try out 8081 first and if that's available use that one.

src/commands/start.ts
```diff
@@ -41,11 +52,52 @@ export async function handler(
...
+                  !Number.isNaN(newPortNumber) &&
+                  newPortNumber <= 65535 &&
+                  newPortNumber > 0
+                ) {
+                  return true;
+                }
+                return 'Please enter a port number between 0 and 65535.';
+              },
+            },
+          ]);
```
dkundel, line 86: Can you factor that out into a function? It largely makes it easier to read this logic.

src/commands/start.ts
philnash: We agreed that random was cool, but we clamped it between 1025 and 65536.

PR #132, opened 2020-04-28
feat(logger): expose errors better & upgrade serverless

dkundel:
**Contributing to Twilio**

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -2,6 +2,7 @@ import {
   DeployLocalProjectConfig,
   TwilioServerlessApiClient,
 } from '@twilio-labs/serverless-api';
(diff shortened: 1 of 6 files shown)


src/commands/deploy.ts
```diff
@@ -60,17 +58,9 @@ function handleError(
...
-    ) as ApiErrorResponse;
-    const messageBody = stripIndent`
-      ${responseBody.message}
-
-      More info: ${responseBody.more_info}
-    `;
-
-    logger.error(messageBody);
+  } else if (err.name === 'TwilioApiError') {
+    const apiError = err as ClientApiError;
```
philnash, line 62: Why do you need to cast this error as `ClientApiError` but in the activate, list and logs commands you don't?

PR #141, opened 2020-06-09
feat(runtime): add CORS headers to OPTIONS requests to assets

### LINKED ISSUE #107: Improvement: CORS headers
I'd be great if CORS headers returned by the local server were set to the same values used by Twilio Runtime once deployed:

```
access-control-allow-origin: *
access-control-allow-methods: GET, PUT
access-control-max-age: 3000
```

This would make for a more consistent testing experience.

Something like this might even possibly suffice, although I'm not sure if Functions should be treated differently than Assets.

```
--- dist/runtime/server.js
+++ dist/runtime/server.js
@@ -45,6 +45,12 @@
         app.use(express_useragent_1.default.express());
         app.use(body_parser_1.default.urlencoded({ extended: false }));
         app.use(body_parser_1.default.json());
+        app.use(function (req, res, next) {
+            res.header("Access-Control-Allow-Origin", "*");
+            res.header("Access-Control-Allow-Methods", "GET, PUT");
+            res.header("Access-Control-Max-Age", "3000");
+            next();
+        });
         app.get('/favicon.ico', (req, res) => {
             res.redirect('https://www.twilio.com/marketing/bundles/marketing/img/favicons/favicon.ico');
         });
```

philnash:
Fixes #107. 

I checked and when you make an OPTIONS request to an asset you get these CORS headers:

```
access-control-allow-origin: *
access-control-allow-headers: Accept, Authorization, Content-Type, If-Match, If-Modified-Since, If-None-Match, If-Unmodified-Since
access-control-allow-methods: GET, POST, OPTIONS
access-control-expose-headers: ETag
access-control-max-age: 86400
access-control-allow-credentials: true
```

But you don't get those headers for other requests.

**Contributing to Twilio**

### DIFF
--- a/src/runtime/server.ts
+++ b/src/runtime/server.ts
@@ -177,18 +177,18 @@ export async function createServer(
         if (routeInfo.filePath) {
           if (routeInfo.access === 'private') {
             res.status(403).send('This asset has been marked as private');
+          } else if (req.method === 'OPTIONS') {
+            res.set({
+              'access-control-allow-origin': '*',
+              'access-control-allow-headers':
+                'Accept, Authorization, Content-Type, If-Match, If-Modified-Since, If-None-Match, If-Unmodified-Since',
+              'access-control-allow-methods': 'GET, POST, OPTIONS',
+              'access-control-expose-headers': 'ETag',
+              'access-control-max-age': '86400',
+              'access-control-allow-credentials': true,
+            });
+            res.status(200).end();
           } else {
-            if (req.method === 'OPTIONS') {
-              res.set({
-                'access-control-allow-origin': '*',
-                'access-control-allow-headers':
-                  'Accept, Authorization, Content-Type, If-Match, If-Modified-Since, If-None-Match, If-Unmodified-Since',
-                'access-control-allow-methods': 'GET, POST, OPTIONS',
-                'access-control-expose-headers': 'ETag',
-                'access-control-max-age': '86400',
-                'access-control-allow-credentials': true,
-              });
-            }
             res.sendFile(routeInfo.filePath);
           }
         } else {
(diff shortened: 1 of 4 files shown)


src/runtime/server.ts
```diff
@@ -178,6 +178,17 @@ export async function createServer(
...
+            if (req.method === 'OPTIONS') {
+              res.set({
+                'access-control-allow-origin': '*',
+                'access-control-allow-headers':
+                  'Accept, Authorization, Content-Type, If-Match, If-Modified-Since, If-None-Match, If-Unmodified-Since',
+                'access-control-allow-methods': 'GET, POST, OPTIONS',
+                'access-control-expose-headers': 'ETag',
+                'access-control-max-age': '86400',
+                'access-control-allow-credentials': true,
+              });
```
dkundel, line 190: Does the options request actually return the content? if not we should probably return the same payload here.

philnash: I don't believe an `OPTIONS` request to a Twilio Asset returns content, no. 

Do you mean this should avoid returning the content then? That is, it should be more like:

```javascript
        if (routeInfo.filePath) {
          if (routeInfo.access === 'private') {
            res.status(403).send('This asset has been marked as private');
          } else if (req.method === 'OPTIONS') {
            res.set({
              'access-control-allow-origin': '*',
              'access-control-allow-headers': 'Accept, Authorization, Content-Type, If-Match, If-Modified-Since, If-None-Match, If-Unmodified-Since',
              'access-control-allow-methods': 'GET, POST, OPTIONS',
              'access-control-expose-headers': 'ETag',
              'access-control-max-age': '86400',
              'access-control-allow-credentials': true,
            });
            res.status(200).end();
          } else {
            res.sendFile(routeInfo.filePath);
          }
        } else {
          res.status(404).send('Could not find asset');
        }
```

That is, send an empty body to an options request and send the file if it's not options.

dkundel: yeah :)

Issue #135, opened 2020-05-13
callback doesn't always terminate execution when running locally


jandusek:
When running a function locally, calling `callback()` doesn't always terminate the execution. When the function is deployed however, `callback()` seems to terminate execution as expected.

Example function that will always throw an error when ran locally, but will run fine when deployed:
```
exports.handler = function (context, event, callback) {
  let response = new Twilio.Response();
  callback(null, response);
  const notReallyUsed = assets['/html/index.html'].open();
  response.appendHeader('Content-Type', 'text/html');
};
```

dkundel:
Yeah this is a known issue atm. We are currently not sandboxing the functions in any way that would allow us to terminate the execution. We should definitely add that though. We are currently discussing the next major version release so this would probably come with that. But it definitely will take some time. Sorry about that.

zladuric:
Just an extra opinion here, but this would _not_ be the behavior I would've expected. If I wanted to end my function after the callback, I want to return myself, explicitly, and not by some magic.

The reason is, what if I have some sort of cleanup logic after the callback? A message to a queue, a log message, a database write?

I understand that I might want to end the result if it succeeds. But then if I depended on something in the callback, I would wrap that part in try/catch block. (And await the result, if the callback can fail asynchronously).

This is how things work in Node (and other JavaScript engines that I know of). Magical exit would definitely break expectations.

Just my humble thoguths on the topic.

philnash:
@zladuric Thanks for your input here. The thing about this is that we are not dealing just with the behaviour of Node.js, but the behaviour of Twilio Functions. [As it says in the documentation](https://www.twilio.com/docs/runtime/functions/invocation#callback-and-asynchronous-limitations):

> It is important to note that when the callback function is invoked, it will terminate all execution of your Function. This includes any asynchronous processes you may have kicked off during during the execution of your handler.
>
> For this reason, if you are using libraries that are natively asynchronous, you need to nest your call to the callback function inside the promise handler.

This issue is necessary so that we can replicate this behaviour in the local version with `twilio-run` so that you don't expect lines after the callback to be run in the deployed environment because they did in the local one.

I hope this makes sense.

zladuric:
Yep, thanks for the in-depth explanation. Allthough I should then probably take my reasoning to the actual Functions thing, not to the twillio-run. But it's not something that I currently use in any case so I'll leave it in your hands.

philnash:
This has been made default for the next version of `twilio-run` in #238.

PR #147, opened 2020-06-29
feat(runtime): experimental: load functions in a separate process

philnash:
…after callback

WIP for #135

This work replaces `functionToRoute` code that picks a function, creates the context and environment for it, runs it and responds to the express application with `functionPathToRoute` that forks a child process to do the above, killing the process once the callback is called.

Currently this works and will run and serve functions successfully in local development mode.

Current issues:

* The integration tests don't pass. As far as I can see, it is because the tests are trying to load the `functionRunner` from the TS directory and not the built JS project
* Functions are loaded fresh every time, because they are loaded into a new process. This is a breaking change and invalidates the `--live` flag (as it is always live). This is a potential performance issue.
* I've left some old code lying around, including:
  * setting the twilio client on the context (this is done in the forked process now)
  * the old `functionToRoute` method that was being used to run the function

I'd appreciate a look over what's been done so far and any opinions. I was surprised I managed this over the course of a stream, so I'm sure there are some very rough parts.

**Contributing to Twilio**

### DIFF
--- a/src/config/start.ts
+++ b/src/config/start.ts
@@ -36,6 +36,7 @@ export type StartCliConfig = {
   appName: string;
   assetsFolderName?: string;
   functionsFolderName?: string;
+  forkProcess: boolean;
 };
 
 export type StartCliFlags = Arguments<
@@ -54,6 +55,7 @@ export type StartCliFlags = Arguments<
     legacyMode: boolean;
     assetsFolder?: string;
     functionsFolder?: string;
+    forkProcess: boolean;
   }
 >;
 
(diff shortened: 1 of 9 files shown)


src/config/start.ts
```diff
@@ -54,6 +55,7 @@ export type StartCliFlags = Arguments<
     legacyMode: boolean;
     assetsFolder?: string;
     functionsFolder?: string;
+    forkProcess: boolean;
```
dkundel, line 58: could we prefix this with `experimental`? Like `--experimental-fork-process` it follows the same schema that Node.js uses for experimental features and it sets the expectation. Because that way people shouldn't be surprised if this flag later disappears

philnash: Good idea, didn't think about making it obviously experimental.

PR #150, opened 2020-07-06
fix(runtime): delete application files from require cache on page load in live mode

### LINKED ISSUE #108: Clean require cache between requests during `live` mode
Right now when you run `--live` mode (default) we are invalidating the `require` cache for functions to make sure they are up-to-date. However, if you `require` a private asset from within a Function these are not being updated on subsequent requests. For example:

```js
const demo = require(Runtime.getAssets()['/demo'].path);
```

There's two options we could fix this.

Option 1) 
Clear the entire cache before every request if `live` mode is enabled.

Option 2)
Use the existing file watcher that has been added that is watching the `assets` and `functions` directory and clear the require cache for that specific file that has been updated.

philnash:
This middleware will delete items from the cache that are not native extensions (cache key ends in ".node") or part of the twilio-run platform on every request when the server is in live mode.

Fixes #108.

**Contributing to Twilio**

### DIFF
--- a/src/runtime/server.ts
+++ b/src/runtime/server.ts
@@ -26,21 +26,22 @@ const DEFAULT_PORT = process.env.PORT || 3000;
 const RELOAD_DEBOUNCE_MS = 250;
 const DEFAULT_BODY_SIZE_LAMBDA = '6mb';
 
-function requireUncached(module: string): any {
-  delete require.cache[require.resolve(module)];
-  return require(module);
+function loadTwilioFunction(fnPath: string): ServerlessFunctionSignature {
+  return require(fnPath).handler;
 }
 
-function loadTwilioFunction(
-  fnPath: string,
-  config: StartCliConfig
-): ServerlessFunctionSignature {
-  if (config.live) {
-    debug('Uncached loading of %s', fnPath);
-    return requireUncached(fnPath).handler;
-  } else {
-    return require(fnPath).handler;
-  }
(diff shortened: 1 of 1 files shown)


src/runtime/server.ts
```diff
@@ -26,21 +26,22 @@ const DEFAULT_PORT = process.env.PORT || 3000;
...
-    return require(fnPath).handler;
-  }
+function requireCacheCleaner(
+  req: ExpressRequest,
+  res: ExpressResponse,
+  next: NextFunction
+) {
+  debug('Deleting require cache');
+  Object.keys(require.cache).forEach(key => {
+    if (!(key.endsWith('.node') || key.includes('twilio-run'))) {
```
dkundel, line 40: what are the things that end with `.node`? :)

dkundel: Could you add a comment here just to explain why we would want to keep those?

dkundel (changes requested):

src/runtime/server.ts
philnash: I will add a comment, yes. But things that end in `.node` are native extensions and as I was looking into this discovered plenty of places that said that unloading native extensions from the cache has undefined behaviour and could cause problems.

philnash:
This is failing with a TS compilation error that I can't reproduce 😖

philnash:
Closing to review conflicts and create new PR on monorepo.

Issue #31, opened 2019-07-12
Protected functions should be protected locally


philnash:
When starting to test the support for protected functions, I expected that I would receive a 401 response when calling my protected function locally without a signature.

I understand that when deploying protected functions will be protected, but to make that more obvious when developing, I think that protected functions should still be protected. This will give confidence to developers when testing functions with webhooks and ngrok.

To make for easier testing, we could add a flag that _unprotects_ protected functions locally.

dkundel:
So here's the issue I found with this one:
If someone uses the Twilio CLI to initialize the project we don't actually have the Auth Token to verify the signature. We could send them a warning in the case that `ACCOUNT_SID` and `AUTH_TOKEN` are actually `API_KEY` and `API_SECRET`. This should probably go into a bigger discussion in a separate issue about how we handle that case. Because there's another situation that is impacted by this.

SpicyPete:
How does one create a `protected` function with this tool?  
I see in the docs that it's possible, but it does not outline how in any way.

philnash:
@SpicyPete To make your function protected you need to name it `function-name.protected.js`. Then, when it is deployed it will require the incoming request to be signed with a `X-Twilio-Signature` header.

This issue just points out that when running locally, protected functions are not protected. So protection is only manifested when deployed.

SpicyPete:
@philnash Great thank you, that is useful to know.  
Is there anywhere in the documentation in the documentation which mentions how this is done where I can read further?  I've been looking everywhere

philnash:
@SpicyPete There's a bit on it here: https://www.twilio.com/docs/labs/serverless-toolkit/general-usage#functions. 

But we could definitely surface that elsewhere too to make things easier.

PR #151, opened 2020-07-06
docs: improves the usage section of the README

philnash:
This adds some extra usage including conventions around private/protected functions (as #31 pointed out weren't obvious) and other bits too.

**Contributing to Twilio**

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,32 +5,36 @@
 <a href="https://travis-ci.com/twilio-labs/twilio-run"><img alt="Travis (.com)" src="https://img.shields.io/travis/com/twilio-labs/twilio-run.svg?style=flat-square"></a>
 <hr>
 
-* [About](#about)
-* [Installation](#installation)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,18 +58,68 @@ npx twilio-run
 
 ## Usage
 
+Check out the [commands](#commands) for in depth usage, but here are some things you will want to know:
```
dkundel, line 61: Could we link somewhere to the rest of the docs on twilio.com/docs/labs/serverless-toolkit? We'll probably have Maria soon update more of the docs there so it would be useful to have those docs over there as well.

philnash: There is a link to the Twilio docs above, under the headline "About". I have started more docs here though, as the repo is one place that people specifically look for documentation. Being able to get the answers here without an additional click is a better experience (not that twilio.com shouldn't host more in depth documentation about this, which I welcome).

PR #177, opened 2020-08-24
feat(twilio-run:start): options for ngrok config and named tunnel

### LINKED ISSUE #158: Allow ngrok configuration to be passed in
Hi, 

It would be good to be able to add additional options to allow ngrok configuration to be passed in. The following use cases for `--ngrok` action

1. pass in subdomain into ngrok so it generate a consistent URI for ngrok
2. pass in user specified configuration file instead of using ~/.ngrok2/ngrok.yml for configration. This will allow project specific ngrok configuration to be set.

philnash:
This allows a user to provide a named tunnel and/or a custom ngrok config when starting `twilio-run` with the `--ngrok` flag.

These options are complementary but also work individually, like so:

`--ngrok-config=path/to/config`

This attempts to load and parse the file at the path supplied. If it does load, then the settings from the config, most importantly the `authtoken` are used in the tunnel config that is passed to `ngrok.connect`.

If the config file cannot be found at the path it will log a warning message and fallback to trying to find the config file in the default location (`~/.ngrok2/ngrok.yml`) and if that doesn't exist the command will continue as if there is no custom config at all.

`--ngrok-name=tunnelName`

Once the config file loading above has happened, including this flag will attempt to find a tunnel config in the config with the supplied name.

If the tunnel config is found, then the `addr` and `proto` settings are over-written (there's no point in tunneling over a different protocol or to a different port number). If a subdomain is passed in then that is over-written too.

If the tunnel config is not found then we log a warning and fallback to the default config.

---

Let me know if you think the fallbacks make sense, or if a tunnel or config file can't be found it should error out? I'd also like feedback on the flag names.

Before merging the flags should be added to `plugin-serverless` too.

Fixes #158.

**Contributing to Twilio**

### DIFF
--- a/packages/twilio-run/src/config/start.ts
+++ b/packages/twilio-run/src/config/start.ts
@@ -1,9 +1,11 @@
 import { EnvironmentVariables } from '@twilio-labs/serverless-api';
 import dotenv from 'dotenv';
-import { readFileSync } from 'fs';
+import { readFile } from 'fs';
+import { promisify } from 'util';
+const readFilePromise = promisify(readFile);
 import path, { resolve, join } from 'path';
 import { homedir } from 'os';
-import { Arguments, config } from 'yargs';
+import { Arguments } from 'yargs';
 import { ExternalCliOptions, SharedFlags } from '../commands/shared';
 import { CliInfo } from '../commands/types';
 import { EnvironmentVariablesWithAuth } from '../types/generic';
@@ -75,9 +77,9 @@ export async function getUrl(cli: StartCliFlags, port: string | number) {
     if (typeof cli.ngrokConfig === 'string') {
       // If we set a config path then try to load that config. If the config
       // fails to load then we'll try to load the default config instead.
-      const configPath = join(process.cwd(), cli.ngrokConfig);
+      const configPath = join(cli.cwd || process.cwd(), cli.ngrokConfig);
       try {
-        ngrokConfig = parse(readFileSync(configPath, 'utf-8'));
+        ngrokConfig = parse(await readFilePromise(configPath, 'utf-8'));
       } catch (err) {
         logger.warn(`Could not find ngrok config file at ${configPath}`);
       }
@@ -87,7 +89,7 @@ export async function getUrl(cli: StartCliFlags, port: string | number) {
       // `ngrokConfig` to be an empty object.
       const configPath = join(homedir(), '.ngrok2', 'ngrok.yml');
       try {
-        ngrokConfig = parse(readFileSync(configPath, 'utf-8'));
+        ngrokConfig = parse(await readFilePromise(configPath, 'utf-8'));
       } catch (err) {
         ngrokConfig = {};
       }
@@ -167,7 +169,7 @@ export async function getEnvironment(
   if (await fileExists(fullEnvPath)) {
     try {
       debug(`Read .env file at "%s"`, fullEnvPath);
-      const envContent = readFileSync(fullEnvPath, 'utf8');
+      const envContent = await readFilePromise(fullEnvPath, 'utf8');
       const envValues = dotenv.parse(envContent);
       for (const [key, val] of Object.entries(envValues)) {
         env[key] = val;
--- a/packages/twilio-run/src/config/start.ts
+++ b/packages/twilio-run/src/config/start.ts
@@ -13,7 +13,7 @@ import { fileExists } from '../utils/fs';
 import { getDebugFunction, logger } from '../utils/logger';
 import { readSpecializedConfig } from './global';
 import { mergeFlagsAndConfig } from './utils/mergeFlagsAndConfig';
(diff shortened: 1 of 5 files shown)


packages/twilio-run/src/config/start.ts
```diff
@@ -67,12 +67,73 @@ export async function getUrl(cli: StartCliFlags, port: string | number) {
...
-    const ngrokConfig: NgrokConfig = { addr: port };
+    // Setup default ngrok config, setting the protocol and the port number to
+    // forward to.
+    const defaultConfig: INgrokOptions = { addr: port, proto: 'http' };
+    let tunnelConfig = defaultConfig;
+    let ngrokConfig;
+    if (typeof cli.ngrokConfig === 'string') {
+      // If we set a config path then try to load that config. If the config
+      // fails to load then we'll try to load the default config instead.
+      const configPath = join(process.cwd(), cli.ngrokConfig);
```
dkundel: should this be `cli.cwd || process.cwd()`?

packages/twilio-run/src/config/start.ts
```diff
@@ -67,12 +67,73 @@ export async function getUrl(cli: StartCliFlags, port: string | number) {
...
+      } catch (err) {
+        logger.warn(`Could not find ngrok config file at ${configPath}`);
+      }
+    }
+    if (!ngrokConfig) {
+      // Try to load default config. If there is no default config file, set
+      // `ngrokConfig` to be an empty object.
+      const configPath = join(homedir(), '.ngrok2', 'ngrok.yml');
+      try {
+        ngrokConfig = parse(readFileSync(configPath, 'utf-8'));
```
dkundel: We are already in an async function. Could we turn this into an async file read?

dkundel (changes requested):
So just that I understand the difference between `--ngrok=dk` and `--ngrok-name=dk` both will be valid with this change but `--ngrok=dk` will create a tunnel under `dk.ngrok.io` while `--ngrok-name=dk` will be based on a name that I have in my ngrok tunnel? Could we add some links to the ngrok docs for named tunnels to avoid confusion?
Also what happens if a named tunnel has a fixed port number? shouldn't we override that? That otherwise the tunnel would be established but it wouldn't actually forward to this

packages/twilio-run/src/config/start.ts
philnash: Ah, forgot the cli could set that. You're right.

packages/twilio-run/src/config/start.ts
philnash: Ha, ok. `readFileSync` had already been imported at this point, but happy to use the async version.

philnash:
> So just that I understand the difference between `--ngrok=dk` and `--ngrok-name=dk` both will be valid with this change but `--ngrok=dk` will create a tunnel under `dk.ngrok.io` while `--ngrok-name=dk` will be based on a name that I have in my ngrok tunnel? Could we add some links to the ngrok docs for named tunnels to avoid confusion?

I can definitely add links to the ngrok docs, but this is correct.

> Also what happens if a named tunnel has a fixed port number? shouldn't we override that? That otherwise the tunnel would be established but it wouldn't actually forward to this

Currently this code over-rides port number and protocol type from the config (line 102). That's why I left a bunch of comments throughout this, but this does need to be documented for end users better too.

packages/twilio-run/src/config/start.ts
```diff
@@ -67,12 +69,73 @@ export async function getUrl(cli: StartCliFlags, port: string | number) {
...
     if (typeof cli.ngrok === 'string' && cli.ngrok.length > 0) {
-      ngrokConfig.subdomain = cli.ngrok;
+      // If we've asked for a custom subdomain, override the tunnel config with
+      // it.
+      tunnelConfig.subdomain = cli.ngrok;
+    }
+    const ngrok = require('ngrok');
+    try {
+      // Try to open the ngrok tunnel.
+      url = await ngrok.connect(tunnelConfig);
```
ShelbyZ, line 126: Should we not just use `configPath` option passed into ngrok.connect and avoid trying to manually parse out params for ngrok? We could definitely do a fs.exist(...) on the file to ensure presence, but we should not be required to fallback and load the default ngrok file stored in the home location (that's free by default).

philnash: We want the ngrok tunnel to connect to the http server we're starting on the port we're starting it on. The config options could be trying to setup a tcp tunnel to a different port. So, I chose to parse the config and turn it into something that will work within the context of a `twilio-run` powered application.

Note, even if you create a named tunnel that points to port 3000, if we start `twilio-run` and something else is already running on port 3000 we offer a new random port number to start on instead. Parsing and manipulating the config in this way will result in a correct tunnel in that case, which would be the expected behaviour.

My idea here is that the tunnel we create will always point to the server we are running, even if it overrides the actual config, because that is what the user would expect to happen.

PR #178, opened 2020-08-25
fix(serverless-api): retries on 429 error for POST request

philnash:
Got by default does not retry on POST requests. This changes the retry behaviour to retry across all of the methods that the API uses (GET, POST and DELETE) but only for 429 errors.

Excuse all the other changes, I think they came from the linter during commit. The important lines (right at the bottom) are: 

```diff
    options.retry = {
      limit: this.config.retryLimit || RETRY_LIMIT,
+     methods: ["GET", "POST", "DELETE"],
+     statusCodes: [429]
    };
```

**Contributing to Twilio**

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,7 @@
+{
+  "trailingComma": "es5",
+  "tabWidth": 2,
+  "semi": true,
+  "singleQuote": true,
+  "parser": "typescript"
+}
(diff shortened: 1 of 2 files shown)


dkundel (changes requested):
I think the prettier config might have gone lost during the move. Can you re add it to the top level (https://github.com/twilio-labs/serverless-toolkit/blob/master/.prettierrc) and fix the formatting? I'll merge this in then :)

philnash:
Added prettierrc, squashed and force pushed up. The diff is better now. Did you have opinions about brackets around single arguments to arrow functions though? 😄

dkundel (approved):

PR #192, opened 2020-09-21
Fix log tailing

dkundel:
It appears that the logs endpoint of the Serverless API doesn't always return predictable results, meaning at times we get empty pages on requests, followed by populated ones upon the next one. That might cause the rendering of duplicate logs. While the long term fix should probably be on the API side, this should fix it on the client side.

Instead of clearing the cache of already seen SIDs upon every page request, we'll cache them until we hit a certain defined limit (default 1000). That should reduce the risk of duplicate logs.

**Note:** logs that were executed directly after each other might still appear out of order. This is a limitation of the logging infrastructure.

Since I wasn't able to reproduce the initial issue, I ended up publishing a version of this fix under the canary tag. You can install it through:

```bash
twilio plugins:remove @twilio-labs/plugin-serverless
twilio plugins:install @twilio-labs/plugin-serverless@canary
```

**Contributing to Twilio**

### DIFF
--- a/packages/plugin-serverless/src/utils.js
+++ b/packages/plugin-serverless/src/utils.js
@@ -1,6 +1,6 @@
-const path = require("path");
-const camelCase = require("lodash.camelcase");
-const { flags } = require("@oclif/command");
+const path = require('path');
+const camelCase = require('lodash.camelcase');
+const { flags } = require('@oclif/command');
 
 function convertYargsOptionsToOclifFlags(options) {
   const flagsResult = Object.keys(options).reduce((result, name) => {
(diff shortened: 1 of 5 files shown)


bobtfish:
I can confirm that this change fixes the observed issue for me, thanks!

bobtfish:
Bump on this. Still seeing this problem and asking people in my teams to use an unreleased version to work around isn't a great developer experience.

dkundel:
Hey @bobtfish sorry for the delay. I opened a PR to backport this fix to the current version to unblock you. I'll notify you once this has been released but we'll release this as a patch fix in our downstream dependency. So you might have to reinstall the plugin to receive the changes.

https://github.com/twilio-labs/serverless-api/pull/58

philnash (approved):

dkundel:
@bobtfish I shipped a patch fix to the underlying serverless-api dependency. So you should be able to get those changes when you run:

```
twilio plugins:remove @twilio-labs/plugin-serverless
twilio plugins:install @twilio-labs/plugin-serverless@latest
```

PR #196, opened 2020-10-16
fix(serverless-api): explicitly enable credentials

dkundel:
I had a typo in the IncludeCredentials. This still worked because the default is true but we want to make sure to explicitly set it

**Contributing to Twilio**


philnash (approved):
LGTM

PR #198, opened 2020-10-30
feat(twilio-run): restructure configuration

### LINKED ISSUE #166: Redesigning config file / .twilio-functions file
Together with the #163 this is probably one of the biggest pains and also a very undocumented feature at the same time. I think it's an issue that should be solved as part of the big changes in v3
## Background

Right now in v2 we have a `.twilio-functions` file that solves for two situations at once:
1) It keeps track of the `serviceSid` and `latestBuildSid` deployed by the tool. The reason why we keep track of these is to streamline the experience of certain commands. For more info see the Deployment Info sections further down
2) It allows you to define configuration values for commands without having to pass them as flags to the command. It also supports per environment and per account/project configurations. For more info see Command Configuration

## Problems

- `.twilio-functions` is an odd file in the sense that for some people it makes sense to commit it but for others it doesn't. This gets especially tricky if you want to add a configuration file to your example project that people might deploy and therefore you don't want to store info like `serviceSid` in it.
- The configuration part is not really well documented and could be useful for more advanced users. 
- The way the `serviceSid` right now is used often confuses people resulting in people deleting it and running `twilio serverless:deploy --override-existing-project` without really knowing why.

## Proposal

Instead of one file we should really have two files. One that enables people to configure the behavior of the Serverless Toolkit and one that we use to store deployment info. 

### Deployment Info

I think this goes hand in hand with fixing #163 where we need to get to using Account SID across the board independently of what we use as deployment credentials. 

Once we have #163 solved we should still store info such as `serviceSid` and `latestBuildSid` as they are used to create convenience commands such as `twilio serverless:deploy`, `twilio serverless:promote` or `twilio serverless:list` without having to pass a variety of additional SID variables. 

However, instead of mashing this together with other configuration we'll store this information separately. This file would by default be part of the `.gitignore` file when we create new projects. It could also host other deployment related information when we implement #159 to speed up the process.

The information should be stored per Account SID inside a single file. Therefore if it is in a repository and people deploy to the same account it works but if they deploy to different accounts it will work without throwing an error. Instead it should be handled the same way as when there is no file as part of the project.

If the file does not exist when we start deployment, instead of currently trying to create a service with a name first, we will try to find a service with the name first and if one is found we prompt the user if they want to use that one. If they create a new one we will write this immediately into the new file for the given Account SID to avoid issues if the deployment fails later in the process.

**Important**: Service SIDs can still be overwritten by the configuration or commandline flags.

Proposed file name: `.twilio-serverless.deployinfo` (I'm open for other alternative names)

### Command Configuration

The overall way the configuration works should not really change, instead it would be separated and better documented. It should support any value that's available as a flag as a config value in camelCase.

Configuration should be possible per:
- project (identified using the Account SID)
- environment (identified using the environment suffix)
- command (identified as the subcommand triggered, e.g. `deploy` or `promote`)
- top-level (shared across all commands, projects and environments)

Config files would be found and loaded using the [`cosmiconfig` module](https://www.npmjs.com/package/cosmiconfig) and the `twilio-serverless` namespace. This provides a variety of different ways to define the config per users preference. The preferred default would be `.twilio-serverlessrc`. We would additionally add support for `json5` if someone uses `.twilio-serverlessrc.json` as it allows for comments.

#### Example:

```json5
{
  "logLevel": "verbose",
  "assetsFolder": "dist",
  "projects": {
    "ACxxxxxxxxxxxx": {
      "env": ".env.demo"
    },
    "ACyyyyyyyyyyyy": {
      "env": ".env"
    }
  },
  "environments": {
    // production environment
    "*": {
      "env": ".env.prod"
    },
    "staging": {
      "env": ".env.stage"
    }
  },
  "start": {
    "env": ".env.local"
  }
}
```

This example has intentionally conflicting information. The order of preference is (first gets overwritten by second, etc.)
1. top level
2. project
3. command
4. environment
5. explicit env flag passed in

As an alternative customers can use the `--config` or `-c` flag to specify different config files.

Proposed file name: Powered by cosmiconfig with the namespace `twilio-serverless`.

dkundel:
This PR aims to restructure the way configuration is handled by dropping .twilio-functions files in
favor of a dedicated config file and a deploy info cache file.

BREAKING CHANGE: Drops support for .twilio-functions files and internally restructures activate
files to promote

fix #166

**Contributing to Twilio**


philnash:
A lot of this seems to be the consolidation of flags, which is a nice refactor. Can you point out the parts where the `.twilio-functions` file format has changed to the `.twilioserverlessrc` version? Where is the impact of this change? Thanks!

dkundel:
The big changes are:
1. The `.twilioserverlessrc` file supports various formats incl json5 (default), yml and JS
2. I removed more complex nesting capabilities (previously you could for example specify config per environment per account). Now you can only do either. The previous was more flexible but both harder to parse and harder to explain.
3. The `.twilioserverlessrc` file will never be modified by the Toolkit. Instead the build info gets written into a `.twiliodeployinfo` file on a per account basis. It's the first step on removing some of the pain points we've seen when working with multiple Twilio accounts on the same project. The `.twilioserverlessrc` file can be checked in but the `.twiliodeployinfo` file should be added to gitignore by default (will add that at a later point)


And yes I consolidated all flags. That should hopefully help us also use flags in a more consistent matter.

packages/twilio-run/src/templating/defaultConfig.ts
```diff
@@ -0,0 +1,68 @@
...
+    .map(templateFlagAsConfig)
+    .join('\n');
+  return [
+    '{',
+    `\t"commands": {},`,
+    `\t"environments": {},`,
+    `\t"projects": {},`,
+    lines,
+    '}',
+  ].join('\n');
```
philnash: Any reason to not use `JSON.stringify` here?

philnash:
Splitting this to the two files, publicly shareable and local data, is definitely going to be a good thing.

You were right in your comment about this being a difficult part of the code base! I can't see anything wrong, but you know it better than me.

You have this as draft. Was there more work to be done?

packages/twilio-run/src/templating/defaultConfig.ts
dkundel: Because I'm adding comments since it's JSON5

dkundel:
I'm still adding tests and I want to have a graceful migration option as well.

packages/twilio-run/src/templating/defaultConfig.ts
philnash: Ah, in `templateFlagAsConfig`, got you. Shame JSON5 doesn't make it easier to actually work with the comments it permits.

packages/twilio-run/bin/upgrade-config.js
```diff
@@ -0,0 +1,130 @@
...
+  );
+
+  if (!fileExists(oldConfigPath)) {
+    console.error('Could not find an old config file at "%s"', oldConfigPath);
+    return 1;
+  }
+
+  const oldConfigContent = JSON.parse(await readFile(oldConfigPath, 'utf8'));
+
+  if (!oldConfigContent) {
```
philnash, line 32: Won't JSON parse throw an error rather than return `null`/`undefined`?

packages/twilio-run/src/checks/legacy-config.ts
```diff
@@ -0,0 +1,31 @@
...
+import path from 'path';
+import { fileExistsSync } from '../utils/fs';
+import { logger } from '../utils/logger';
+
+export function printConfigWarning() {
+  const title = 'Legacy Configuration Detected';
+  const msg = stripIndent`
+    We found a .twilio-functions file in your project. This file is incompatible with the current version of the CLI you are using and will be ignored.
+
+    In most cases you will be able to delete the existing .twilio-functions file. If you have any configuration/modifications you did to the file yourself, head over to twil.io/serverlessv3 to learn how to migrate your configuration.
```
philnash: Can you add an `https://` to the URL here? Then in certain terminals (hyper, VSCode) you can command click it.

packages/twilio-run/__tests__/config/utils/mergeFlagsAndConfig.test.ts
```diff
@@ -3,7 +3,7 @@ import '../../../src/config/utils/mergeFlagsAndConfig';
 import { mergeFlagsAndConfig } from '../../../src/config/utils/mergeFlagsAndConfig';
 
 const baseFlags = {
-  config: '.twilio-functions',
+  config: '.twilioserverlessrc.js',
```
philnash: Everywhere else, except in this test, the config is referred to as `.twilioserverlessrc` without the `.js`. Is that an issue?

philnash (commented):
I have a couple of questions. I'm sure nothing is a blocker.

packages/twilio-run/bin/upgrade-config.js
dkundel: Good point! I'll get that fixed

packages/twilio-run/__tests__/config/utils/mergeFlagsAndConfig.test.ts
dkundel: No. The parser should support `.js`, `.json`, `.json5`, etc. I can change to for consistency though

philnash (approved):
:+1:

Issue #205, opened 2020-12-07
More a question about using twilio-run without dependency on ngrok


alsitebro:
Hi,

We currently use the serverless-toolkit for building and running our serverless functions over at my workplace. 
I am currently working on building a CI/CD pipeline to automate the build and release cycles for our projects. Unfortunately, we are restricted from using ngrok within our network and traffic to its services are actively blocked by our firewall, because, security.

As a result, we are unable to install the required npm packages for twilio-run which depends on ngrok.

Is there an option within the serverless toolkit for us to build our pipeline without relying on ngrok. If this is not possible, then it means we may need to create a custom solution.

Could please point me in the direction of documentation which outlines to what extent the ngrok package is required in the toolkit. 

Thanks

philnash:
Hey @alsitebro, thanks for opening this issue.

For the toolkit itself, ngrok is used when starting the local development server in `twilio-run`. This is something we could push to an optional dependency (though I'm unclear whether that would help when installing as a CLI plugin). I will investigate this.

We've also investigated the idea of updating the ngrok package to handle the ngrok executable failing to download, though work on that hasn't started yet either.

For the meantime, you may be able to rely directly on the [@twilio-labs/serverless-api](https://www.npmjs.com/package/@twilio-labs/serverless-api) package, which handles the majority of the deployment logic. The `twilio-run` deploy command does a bunch of collecting data and checking that the project is all correct, then [uses a `TwilioServerlessApiClient` to make the actual deploy](https://github.com/twilio-labs/serverless-toolkit/blob/main/packages/twilio-run/src/commands/deploy.ts#L107-L127).

Does that help at all?

alsitebro:
Thanks for this @philnash , I'll try the serverless-api package. 
And I appreciate you implementing the change to make ngrok optional. That will be really useful.
I am happy for this issue to be closed

PR #207, opened 2020-12-08
feat(twilio-run:start): make ngrok optional and handle failure

philnash:
This makes ngrok an optional dependency and handles the error when loading it by shutting down the server and printing a message.

This is intended to help with #205. Still need to investigate installing CLI plugins without optional dependencies.

**Contributing to Twilio**


philnash:
The tests are failing because Express can't be found? That has nothing to do with my changes and the tests are passing locally. Any ideas? 🤷‍♂️

PR #217, opened 2021-02-12
feat(create-twilio-function): adds twilio to default package.json

### LINKED ISSUE #200: Install latest Twilio Node.js lib by default
When creating a new project we should install the latest Twilio version rather than a hard-coded one.

My suggestion is to either use pkg-install for it and actually install twilio@latest or contact npm during package.json creation to get the latest version. I think this is preferable over specifying * as version since it would require devs to pin the dependency later themselves.

As part of this we should ensure that when a template from function-templates is used that already specifies a specific twilio version (like the verify template), the clash is handled. In this scenario likely the later version is appropriate (npm.im/semver can help).

philnash:
When installing `twilio-run` as a devDependency of a local function repo you get the latest version of `twilio` installed. Deploying that function will then use the default version of `twilio` in Functions which is falling further and further behind. This update adds `twilio` as a direct dependency and installs the latest version allowing developers to see the dependency, understand the version and use their functions locally and deployed with confidence.

**Contributing to Twilio**

### DIFF
--- a/packages/create-twilio-function/src/create-twilio-function/versions.js
+++ b/packages/create-twilio-function/src/create-twilio-function/versions.js
@@ -1,4 +1,5 @@
 module.exports = {
+  twilio: '^3.56',
   twilioRun: '^2.6.0',
   node: '10',
   typescript: '^3.8',
--- a/packages/create-twilio-function/src/create-twilio-function/create-files.js
+++ b/packages/create-twilio-function/src/create-twilio-function/create-files.js
@@ -19,8 +19,11 @@ async function createFile(fullPath, content) {
   return writeFile(fullPath, content, { flag: 'wx' });
 }
 
-const javaScriptDeps = {};
-const typescriptDeps = { '@twilio-labs/serverless-runtime-types': versions.serverlessRuntimeTypes };
+const javaScriptDeps = { twilio: versions.twilio };
+const typescriptDeps = {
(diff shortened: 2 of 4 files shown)


packages/create-twilio-function/src/create-twilio-function/versions.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
+  twilio: '^3.56',
```
dkundel, line 2: Do we want to be explicit here? Or should we just open it up to `^3` or even `*`? I think even if we put an explicit number here people might still get confused.

dkundel (changes requested):

packages/create-twilio-function/src/create-twilio-function/versions.js
philnash: Given that we are placing this into a package.json for a project, leaving it as `*` or `^3` will mean that in the case of `twilio` version 4 being released a fresh `npm install` in that project will pick the new version and potentially cause breaking changes.

I think that `^3.56` is the right level of semver specificity for `create-twilio-function` to install the latest version of `twilio` into a project until we need to change it to version 4.

philnash:
This was backported to `create-twilio-function` on the [`v2` branch](https://github.com/twilio-labs/create-twilio-function/tree/v2).

dkundel (approved):

PR #230, opened 2021-03-25
WIP: Assets plugin

philnash:
This is a new Twilio CLI plugin that can create a new Twilio Runtime Service to be used as a bucket to store assets. You can then upload assets to the bucket and list the currently available assets.

I am starting a draft PR to get feedback on the code and on how is best to release this. Should we get it out as a version 1 alongside the major release of all other serverless toolkit packages?

It includes some changes to serverless-api, though these are really just exporting a couple of functions that weren't previously exported.

**Contributing to Twilio**


dkundel:
I think this looks great overall and I'm happy to release it alongside the rest. The only thing I'd say is that because it's not a standalone like `twilio-run` you should use the built-in Twilio CLI output and logging mechanisms (I think using spinners makes sense) so that `-l` for log levels and `-o` for machine readable output works. I think especially for this plugin and the `assets:list` command having machine readable output is useful.

philnash:
Ah yes, I did mean to include those bits.

I'm actually holding off including this until [twilio-cli-core#118](https://github.com/twilio/twilio-cli-core/pull/118) is deployed with the CLI this week. That way I can dump the `configStore` and use the new built in storage commands in the CLI (and throw an error with a message to upgrade the Twilio CLI if someone tries to use this with an older version).

I meant to add tests too 😄 I can get to them after v1 though.

philnash:
I think this is ready for a real review now. It works with all the basic Twilio CLI flags and while it does still lack tests, it works!

packages/plugin-assets/src/init.js
```diff
@@ -0,0 +1,78 @@
...
+} = require('@twilio-labs/serverless-api/dist/api/services');
+const {
+  createEnvironmentFromSuffix,
+  getEnvironment,
+} = require('@twilio-labs/serverless-api/dist/api/environments');
+const { TwilioCliError } = require('@twilio/cli-core').services.error;
+
+const DEFAULT_ASSET_SERVICE_NAME = 'CLI-Assets-Bucket';
+
+const createServiceAndEnvironment = async client => {
```
dkundel: Any reason why you use arrow functions instead of regular async function declarations?

packages/plugin-assets/src/init.js
```diff
@@ -0,0 +1,78 @@
...
+      const environment = await getEnvironment(
+        environmentSid,
+        serviceSid,
+        client
+      );
+      return environment;
+    } catch (error) {
+      logger.debug(error.toString());
+      throw new TwilioCliError(
+        `Could not fetch asset service environment with config:
```
dkundel: It would be useful to provide some steps here as to why / how to fix it

packages/plugin-assets/src/upload.js
```diff
@@ -0,0 +1,253 @@
...
+
+const { resolve, basename } = require('path');
+const { readFile } = require('fs').promises;
+const EventEmitter = require('events');
+
+const {
+  listFunctionResources,
+} = require('@twilio-labs/serverless-api/dist/api/functions');
+
+const spinner = ora();
```
dkundel: The spinner shouldn't show if you pass `-l error` for example.

packages/plugin-assets/src/upload.js
```diff
@@ -0,0 +1,253 @@
...
+          type: 'list',
+          name: 'conflict',
+          choices: ['Overwrite', 'Rename', 'Abort'],
+        },
+      ]);
+      if (answers.conflict === 'Overwrite') {
+        newAsset.sid = existingAsset.sid;
+        spinner.start('Overwriting existing asset');
+      } else if (answers.conflict === 'Rename') {
+        let newNameAnswers = await inquirer.prompt([
```
dkundel: We should do verification here that it doesn't contain invalid characters

packages/plugin-assets/src/upload.js
```diff
@@ -0,0 +1,253 @@
...
+const { readFile } = require('fs').promises;
+const EventEmitter = require('events');
+
+const {
+  listFunctionResources,
+} = require('@twilio-labs/serverless-api/dist/api/functions');
+
+const spinner = ora();
+
+const upload = async ({
```
dkundel: Can you break the different steps in this function into individual functions? Just makes reading this and debugging it easier.

dkundel (changes requested):

packages/plugin-assets/src/upload.js
philnash: [Invalid characters are](https://www.twilio.com/docs/runtime/functions-assets-api/api/asset-version#create-an-asset-version-resource) `;,?:@+&$()' "` and fragments, so `#`.

Most of these can actually be uri encoded and still work fine. Rather than confuse the code (we should also check filenames which are passed in as the original file, for example) for an edge case I would prefer for now to fail the asset version upload with a better error message (I've added the asset path to the failure message).

We can add this sort of checking in later, if required.

packages/plugin-assets/src/upload.js
philnash: Good point, they are info messages so they should only show for info or debug really. Will update.

packages/plugin-assets/src/init.js
philnash: No reason, have updated to async function declarations.

packages/plugin-assets/src/init.js
philnash: I've added a message to run in debug mode for more detail. Many of the errors throughout these scripts should not occur unless there is a network error or the user has edited the config file by hand and changed the SIDs.

PR #233, opened 2021-04-15
fix(twilio-run:checks): change to Node.js v12 check

dkundel:
I'm bumping the version to Node.js 12. We should find a proper way to detect the Node.js version that is intended probably via nvmrc or engines and then pass that via the `Runtime` flag to the API but for now the warning for 12 seems better.

**Contributing to Twilio**


philnash (approved):
👍

PR #235, opened 2021-04-16
feat(templating): add support for .env.example files

### LINKED ISSUE #112: [discussion] Support `.env.example` for Function templates
Right now github.com/twilio-labs/function-templates uses `.env` files for each template to specify which environment variables are needed. @thorsten-stripe pointed out correctly that that might be dangerous as people might push secrets accidentally. We could switch this to `.env.example` files but I'm concerned how we could do a transition where we don't break older versions of this tool.

dkundel:
This adds support for `.env.example` files in templates as well as continuing to support `.env` for backwards compatibility.

There could be the situation that both exist in which case it would append the content of both. 

Additionally I fixed the fact that the added content didn't contain the namespace name.

fixes #112 

**Contributing to Twilio**


philnash (approved):
:+1:

PR #238, opened 2021-04-16
feat(twilio-run)!: enable process forking by default

### LINKED ISSUE #223: Make running functions in a separate process default

dkundel:
This change enables the process forking by default and removes the "experimental" prefix from the
flag. From now on by default Functions in local development will get executed in a forked process.
This results in no ability to share state between Function invokations. While that is possible in
deployed Functions it's unpredictable and we intend to discourage it this way. If someone has to
rely on that feature they can use --no-fork-process.

BREAKING CHANGE: Functions invokations are not separated from each other

fix #223 #135

**Contributing to Twilio**


philnash (approved):
:+1:

PR #240, opened 2021-04-16
chore(runtime-handler): move runtime-handler into monorepo

dkundel:
Right now the `@twilio/runtime-handler` package wasn't version controlled. It's currently empty but with the next version we'll start extracting the local dev logic from `twilio-run` into this package so tracking them together even though it's a different "scope" makes sense to me.

**Contributing to Twilio**


philnash (approved):
:+1:

PR #246, opened 2021-04-20
feat: check for legacy config

dkundel:
For all "destructive" commands I also added a prompt that allows you to stop execution. Since "start" isn't destructive it only puts a warning out.

Screen_Shot_2021-04-20_at_15_31_44

**Contributing to Twilio**


philnash:
When you say "destructive" do you just mean "hits the API"? Commands like `list` or `logs` don't change things, but do hit the API, but require a confirmation with this change.

dkundel:
Good point! I removed the prompt from list and logs.

philnash (approved):

PR #252, opened 2021-04-30
feat: extract runtime-handler and lazyLoading

### LINKED ISSUE #250: Migrate runtime out of twilio-run
## Background

The Functions team has recently started exposing the runtime version via the `@twilio/runtime-handler` package in order to add new features in the future. In order to create consistency between Functions and local development we need to start adding those features as well and leverage the npm-published version of `@twilio/runtime-handler` that has already been added to this. 

### 1. "Cannot find module"
Today you cannot `require` the runtime handler from inside a Function. For example if you run:

```js
const demo = require('@twilio/runtime-handler');
```

It will return the error message: `Cannot find module '@twilio/runtime-handler'`.

We should mimic the same behavior for the time being.

### 2. Be flexible for new features

While you cannot require the `@twilio/runtime-handler` package today, this might change in the future if new helper/convenience functions are being added. In an effort to avoid adding more things to the global scope. Hence we should make sure that the main entry of `@twilio/runtime-handler` remains empty for the time being.

### 3. Backwards compatibility

There are two ways we could roll these changes out. One would mean that we do a major version bump as soon as the Functions team releases `1.0.2` and force people to use the latest version of the Serverless Toolkit for any runtime handler version 1.0.2 or newer. This would be a pretty frustrating experience but would cut down on maintenance overhead and complexity.

The alternative would be to be backwards compatible. This approach in my opinion wouldn't require a major version bump. For any runtime-handler of 1.0.2 or newer we'd pull the local implementation from the new package. For everything older we'd load the built-in version inside `twilio-run`. Once `1.0.1` has been deprecated we could then remove the local dev experience completely from inside `twilio-run` in the next major version.

## Implementation Recommendation

We duplicate all existing local dev logic that lives in `twilio-run` today into the `@twilio/runtime-handler` package.

From there we'll leverage [package entry points](https://nodejs.org/docs/latest-v12.x/api/packages.html#packages_package_entry_points) (supported as of Node v12) to export the local dev logic as `@twilio/runtime-handler/dev` (open for other names). 

Inside `twilio-run` we'll check if `@twilio/runtime-handler/dev` can be loaded. If that is the case we'll load the wrapper code from there. If it doesn't exist we'll default to the code inside `twilio-run`. We might additionally want to check if `@twilio/runtime-handler` is explicitly mentioned in the project's `package.json`.

*Unsure*: If we point `@twilio/runtime-handler` package entry point to an invalid path this should give the same experience to customers. If it doesn't work we should inject into the wrapper code logic that explicitly removes `@twilio/runtime-handler` from being importable.

The `init` command should additionally check what the latest stable release of the runtime-handler is and install it as a default when creating a project. That way we will slowly move people away from not specifying a runtime-handler version. Since the `@twilio/runtime-handler` requires an exact version number we'll have to manually fetch it from npm.

## Bonus

Later on we can add an additional package entry point for `@twilio/runtime-handler/test` that can be used by customers to make writing tests easier. Essentially some of the logic we already have in the function-templates repo.

dkundel:
This is still in progress. I still have to verify the functionality and port/add tests. As part of this change the new runtime-handler is now using the `twilio` package that person has installed. 

In the past we were relying on the resolution of Node.js to pick up the right one but it might pick up the wrong one. 

fixes: #250 

**Contributing to Twilio**


dkundel:
This PR will also officially drop Node 10 support

packages/twilio-run/src/runtime/server.ts
```diff
@@ -57,6 +60,99 @@ function requireCacheCleaner(
...
+    server.on('request-log', (logMessage) => {
+      writeOutput(logMessage)
+    });
+    server.on('updated-routes', async (config) => {
+      await printRouteInfo(config);
+    })
+    configureWatcher(config, server);
+    return server.getApp();
+  } catch (err) {
+    debug(
```
philnash, line 153: Do we need to catch this if the version of `twilio-run` depends on `runtime-handler`? What is the situation in which this version of `twilio-run` would not have access to a `LocalDevelopmentServer` from this package, regardless of which version it is?

philnash: And consequently, do we need to keep all the server code in `twilio-run` as well as in `runtime-handler`?

dkundel: it doesn't actually depend on runtime-handler. It's a `devDependency` for the type but the require comes from the customers project not bundled from `twilio-run`. We could do both but I feel like that might introduce more edge cases than just ripping the bandaid off in the next major release and dropping all the code.

dkundel:
The PR should be ready for review but shouldn't be merged yet which is why it's in progress

PR #256, opened 2021-05-03
fix(twilio-run:start): load functions & assets in forked process

philnash:
Due to changes around forking processes to run functions, the functions and assets cache wasn't being populated in the forked process. This slightly refactors how routes are cached by moving repeated calls into a `getRouteMap` function within route-cache and then using that in both the server, to populate the cache for the terminal display, and inside the forked process so that it populates the cache before building the global scope.

**Contributing to Twilio**


dkundel (approved):

dkundel:
Good catch. I'll have to port the same change into #252 later on

PR #268, opened 2021-05-18
fix(plugin-serverless): handle direct username/password input

### LINKED ISSUE #261: Deployment with --username and --password as API Key/Secret fails
If you deploy with --username and --password as API Key and API Secret it fails with a 404 where it grabs the wrong Service SID and tries to access /Environments.

regarding #255

dkundel:
Previously if you'd call `twilio serverless:deploy` or similar commands with `--username` and
`--password` explicitly passed as API Key/Secret, it would both read the Service SID and write the
Service SID in the deployinfo cache for the Account SID that was associated with the default
profile. This change will ignore the default profile/client if username or password are passed.

BREAKING CHANGE: You have to use --username and --password effectively together if you use either

fix #261

**Contributing to Twilio**

### DIFF
--- a/packages/plugin-serverless/src/utils.js
+++ b/packages/plugin-serverless/src/utils.js
@@ -69,6 +69,17 @@ function normalizeFlags(flags, aliasMap) {
 function createExternalCliOptions(flags, twilioClient) {
   const profile = flags.profile;
 
+  if (flags.username || flags.password) {
+    return {
+      username: flags.username,
+      password: flags.password,
+      accountSid: flags.username.startsWith('AC') ? flags.username : undefined,
+      profile: undefined,
+      logLevel: undefined,
+      outputFormat: undefined,
+    };
+  }
+
   return {
     username: twilioClient.username,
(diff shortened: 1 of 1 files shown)


packages/plugin-serverless/src/utils.js
```diff
@@ -69,6 +69,17 @@ function normalizeFlags(flags, aliasMap) {
 function createExternalCliOptions(flags, twilioClient) {
   const profile = flags.profile;
 
+  if (flags.username || flags.password) {
+    return {
+      username: flags.username,
```
ktalebian, line 77: Doesn't this mean `username` or `password` could potentially be null here? Is that okay?

ktalebian (approved):

packages/plugin-serverless/src/utils.js
dkundel: Realistically they wouldn't but I wrote a more proper check

ktalebian (approved):

PR #271, opened 2021-05-20
feat: extract runtime-handler and lazyLoading (#252)

dkundel:
This change introduces the following features:
- It adds the local development logic into the public `@twilio/runtime-handler`
- Implements lazyLoading for the Twilio client instances constructed within `@twilio/runtime-handler` to bring parity with the 1.1.0 release in Functions
- Starts using the customer project's `twilio` package instead of the one from the Serverless Toolkit

**Contributing to Twilio**


packages/runtime-handler/package.json
```diff
@@ -38,14 +42,37 @@
...
     "typescript": "^3.8.3"
   },
   "bugs": {
     "url": "https://github.com/twilio-labs/serverless-toolkit/issues"
+  },
+  "dependencies": {
+    "@twilio-labs/serverless-runtime-types": "2.1.0-rc.0",
+    "@types/express": "4.17.7",
+    "chalk": "^4.1.1",
+    "chokidar": "^3.2.3",
```
philnash: This package doesn't appear to use chokidar or lodash.debounce.

packages/runtime-handler/src/dev-runtime/types.ts
```diff
@@ -0,0 +1,88 @@
+export type SearchConfig = {
```
philnash: `runtime-handler` is now exporting some duplicate types that also live in `serverless-api` or `twilio-run`. Should they be consolidated somewhere? 

Also in `serverless-api`:
`SearchConfig`, `AccessOptions`, `ServerlessResourceConfig`, `ServerlessResourceConfigWithFilePath`

Also in `twilio-run`:
`EnvironmentVariablesWithAuth`

philnash (changes requested):
For the most part, this is a copy of twilio-run functionality to runtime-handler and that all seems fine to me. Bringing the lazy loading option in looks just fine to me.

We are duplicating some types, which I wonder if we could bring together somewhere.

And there are some unused dependencies in runtime-handler, which we could remove before publishing.

Otherwise, good lift here!

packages/runtime-handler/src/dev-runtime/types.ts
dkundel: The problem is the way the dependency chain works. `twilio-run` should rely on `serverless-api` and `runtime-handler` (at least during compile time) but `runtime-handler` shouldn't rely on `twilio-run`. And similarly I didn't want `runtime-handler`to rely on `serverless-api` since `runtime-handler` can be different from user to user during runtime so it should be self-contained in general

dkundel: In fact you'll see that runtime-handler is only a devDependency in twilio-run because I do `import type` as opposed to `import`.

philnash: Could runtime-handler take serverless-api as a devDependency and do `import type` too then?

packages/twilio-run/src/utils/requireFromProject.ts
```diff
@@ -0,0 +1,6 @@
+import { createRequire } from 'module';
+import { join } from 'path';
+
+export function requireFromProject(projectDir: string, packageName: string) {
+  return createRequire(join(projectDir, 'node_modules'))(packageName);
```
philnash: This will still fallback to the version installed by `twilio-run` if there isn't an explicit dependency on `twilio` in the project, right? And that is the intention?

packages/runtime-handler/src/dev-runtime/types.ts
dkundel: Unfortunately not because some of these are part of `ServerConfig` which is exported by `runtime-handler`. Realistically the only thing we could do is to move it into a common dependency that all of them consume. But realistically they are all part of `ServerConfig` which is basically the "contract" for runtime-handler. So I could see scenarios where they could diverge slightly from the other use cases.

I was able to remove `SearchConfig` though since it's not being used anymore.

packages/twilio-run/src/utils/requireFromProject.ts
dkundel: I made the fallback more explicit now.

philnash: Have you pushed? The last commits I can see are from 13 days ago.

dkundel: Sorry I was changing around logic afterwards. It's now up-to-date

philnash (approved):
I think this is looking good!

PR #285, opened 2021-06-23
fix(serverless-api): do not retry for HTTP client request errors

### LINKED ISSUE #284: Setting a region in a profile causes deploy to hang
If you have a Twilio CLI profile with a region set, or try to pass a region to the command, the deploy hangs, seemingly indefinitely.

```
twilio serverless:deploy --region au1
```

The root problem is that `serverless.au1.twilio.com` doesn't exist and the HTTP client is failing with `ENOTFOUND`.

The deploy seems to hang because the retry logic in `got` will retry not just for certain failure status codes (which we restrict to 429 only) but for certain errors (the got docs list `ETIMEDOUT`, `ECONNRESET`, `EADDRINUSE`, `ECONNREFUSED`, `EPIPE`, `ENOTFOUND`, `ENETUNREACH` and `EAI_AGAIN`). The default number of retries is 10 with exponential back off, so the deploy seems to hang because the first retry takes 1 second, then 2 seconds, up to about 8.5 minutes for the final retry.

Retrying for these errors seems like a bad idea as they may not resolve and the back off makes it take too long for a reasonable error message.

I propose that we update the [retry options for the client](https://github.com/twilio-labs/serverless-toolkit/blob/main/packages/serverless-api/src/client.ts#L687-L691) to:

```js
    options.retry = {
      limit: this.config.retryLimit || RETRY_LIMIT,
      methods: ['GET', 'POST', 'DELETE'],
      statusCodes: [429],
      errorCodes: []
    };
```

We should also ensure that the error message for `ENOTFOUND` is useful to a user of serverless-api and to `plugin-serverless`.

Finally, [we support adding a `region` and `edge` to the URL](https://github.com/twilio-labs/serverless-toolkit/blob/main/packages/serverless-api/src/api/utils/api-client.ts) but what should the behaviour be when that region or edge is not supported by the Serverless API?

philnash:
Fixes #284

When you run `twilio serverless:deploy --region au1` with this patch it now fails like this:

Failed deployment. Error. getaddrinfo ENOTFOUND serverless.au1.twilio.com

**Contributing to Twilio**

### DIFF
--- a/packages/serverless-api/src/client.ts
+++ b/packages/serverless-api/src/client.ts
@@ -343,13 +343,8 @@ export class TwilioServerlessApiClient extends events.EventEmitter {
    */
   async activateBuild(activateConfig: ActivateConfig): Promise<ActivateResult> {
     try {
-      let {
-        buildSid,
-        targetEnvironment,
-        serviceSid,
-        sourceEnvironment,
-        env,
-      } = activateConfig;
+      let { buildSid, targetEnvironment, serviceSid, sourceEnvironment, env } =
+        activateConfig;
 
       if (!buildSid && !sourceEnvironment) {
         const error = new Error(
@@ -688,6 +683,7 @@ export class TwilioServerlessApiClient extends events.EventEmitter {
       limit: this.config.retryLimit || RETRY_LIMIT,
       methods: ['GET', 'POST', 'DELETE'],
       statusCodes: [429],
+      errorCodes: [],
     };
     return this.limit(() => this.client[method](path, options));
   }


packages/serverless-api/src/client.ts
```diff
@@ -688,6 +683,7 @@ export class TwilioServerlessApiClient extends events.EventEmitter {
       limit: this.config.retryLimit || RETRY_LIMIT,
       methods: ['GET', 'POST', 'DELETE'],
       statusCodes: [429],
+      errorCodes: [],
```
dkundel, line 686: This is interesting isn't 429 technically an error code?

philnash: The `errorCodes`refer to this list: `ETIMEDOUT`, `ECONNRESET`, `EADDRINUSE`, `ECONNREFUSED`, `EPIPE`, `ENOTFOUND`, `ENETUNREACH` and `EAI_AGAIN`. The sort of errors that you don't get a status code for, because something broke before you get a response.

Come to think of it, it might be useful to be able to configure which errors/statuses that this library will retry. Only retrying 429s is useful for the CLI side of things, but that isn't the only thing this could be used for. I think this PR fixes our initial issue, but could be improved to serve users of `serverless-api` in general.

dkundel (approved):

PR #286, opened 2021-06-24
feat(plugin-assets): adds flag for setting service name in init

philnash:
Also generates random project name to avoid all services being called the same thing.

**Contributing to Twilio**

### DIFF
--- a/packages/plugin-assets/src/commands/assets/init.js
+++ b/packages/plugin-assets/src/commands/assets/init.js
@@ -9,7 +9,6 @@ class InitCommand extends TwilioClientCommand {
   async run() {
     await super.run();
     try {
-      const { flags } = this.parse(InitCommand);
       const pluginConfig = getPluginConfig(this);
       const result = await init({
         apiKey: this.currentProfile.apiKey,
@@ -17,7 +16,7 @@ class InitCommand extends TwilioClientCommand {
         accountSid: this.currentProfile.accountSid,
         pluginConfig: pluginConfig,
         logger: this.logger,
-        serviceName: flags['service-name'],
+        serviceName: this.flags['service-name'],
       });
       this.output(result, this.flags.properties);
     } catch (error) {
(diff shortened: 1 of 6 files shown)


packages/plugin-assets/src/commands/assets/init.js
```diff
@@ -1,19 +1,23 @@
...
 const { getPluginConfig } = require('../../pluginConfig');
+const generateProjectName = require('project-name-generator');
+
 const { init } = require('../../init');
 
 class InitCommand extends TwilioClientCommand {
   async run() {
     await super.run();
     try {
+      const { flags } = this.parse(InitCommand);
```
dkundel: Why do you need this here but then down on line 22 you use `this.flags` as opposed to `flags`?

packages/plugin-assets/src/commands/assets/init.js
```diff
@@ -1,19 +1,23 @@
 const { flags } = require('@oclif/command');
 const { TwilioClientCommand } = require('@twilio/cli-core').baseCommands;
 const { getPluginConfig } = require('../../pluginConfig');
+const generateProjectName = require('project-name-generator');
```
dkundel, line 4: We recently had some issues with name generators generating offensive names. Do we feel comfortable with this one?

philnash: It appears this would have been a problem, but someone already pruned the adjectives and nouns for problematic words/potential combinations in [this pull request](https://github.com/aceakash/project-name-generator/pull/33) and previously in [this pull request](https://github.com/aceakash/project-name-generator/pull/21).

packages/plugin-assets/src/commands/assets/init.js
philnash: Mixed examples in other places I guess! Have updated to always use `this` and not parse for a second time.

dkundel (approved):

PR #287, opened 2021-06-24
fix(response): brings Response object to parity with Functions

philnash:
I was checking the code behind the `Response` object that Functions uses and noticed some differences, such as:

* set/append functions returning `this` so that you can chain functions
* body is initialized as `null` not `undefined`
* being able to set headers, body and status code in the constructor

I think this brings the serverless toolkit version of the `Response` up to parity in both `twilio-run` and `runtime-handler`.

**Contributing to Twilio**

### DIFF
--- a/packages/runtime-handler/__tests__/dev-runtime/internal/response.test.ts
+++ b/packages/runtime-handler/__tests__/dev-runtime/internal/response.test.ts
@@ -3,31 +3,52 @@ import { Response } from '../../../src/dev-runtime/internal/response';
 
 test('has correct defaults', () => {
   const response = new Response();
-  expect(response['body']).toBeUndefined();
+  expect(response['body']).toBeNull();
   expect(response['statusCode']).toBe(200);
   expect(response['headers']).toEqual({});
+});
+
+test('sets status code, body and headers from constructor', () => {
+  const response = new Response({
+    headers: {
+      'Access-Control-Allow-Origin': 'example.com',
+      'Access-Control-Allow-Methods': 'GET,PUT,POST,DELETE',
+      'Access-Control-Allow-Headers': 'Content-Type',
+    },
+    body: 'Error',
+    statusCode: 400,
+  });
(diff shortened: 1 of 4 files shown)


packages/runtime-handler/__tests__/dev-runtime/internal/response.test.ts
```diff
@@ -3,31 +3,52 @@ import { Response } from '../../../src/dev-runtime/internal/response';
...
+    'Access-Control-Allow-Methods': 'GET,PUT,POST,DELETE',
+    'Access-Control-Allow-Headers': 'Content-Type',
+  });
+});
+
 test('sets status code', () => {
   const response = new Response();
   expect(response['statusCode']).toBe(200);
-  response.setStatusCode(418);
+  const response2 = response.setStatusCode(418);
```
dkundel: Shouldn't the better test here be to re-check `response` because technically even though it returns itself we want to make sure it updated the original response object? This test would still succeed if we accidentally changed the logic where setStatusCode returns a new object

dkundel: and then have a dedicated test that calling the same function twice chained returns the same instance.

This really applies throughout the PR

dkundel (changes requested):

packages/runtime-handler/__tests__/dev-runtime/internal/response.test.ts
philnash: I agree that this should have been a different test, that was lazy on my behalf 😄. I've updated that now.

I don't think we need to call this twice and test after though. Using `expect(object1).toBe(object2)` checks referential identity of the objects. So if we started returning a new object from `setStatusCode`, or the other functions here, this test would fail. Then the existing tests ensure that the changes are correctly made on the original object.

dkundel (approved):

Issue #249, opened 2021-04-26
Provide a way to update variables


dkundel:
Right now there is no easy way to update environment variables using the Serverless Toolkit. You can do so using `twilio api:serverless:...` but it requires the knowledge of the Variable SID and other information.

We should provide a way to set those variables more conveniently. 

## Suggestion

Provide a dedicated command `variables` to set, get and update environment variables without having to know whether they exist or not. 

### Update an environment variable for `dev` environment from within a project using the Twilio CLI

```
twilio serverless:env set --key NAME --value VALUE
```

### Update environment variables based on a `.env` file in production

```
twilio serverless:env import --env .env.prod --production
```

### Get variables from a `stage` environment

```
twilio serverless:env get --key NAME --environment stage
```

### Remove a variable from a specific environment using the environment SID

```
twilio serverless:env unset --key DEBUG --environment ZE1222222222222222222222
```

### List all environment variables

```
twilio serverless:env list --show-values --service-sid ZS11111111111111111111111 --environment ZE333333333333333
```

## Inspiration

The terms are inspired by the [Netlify CLI](https://cli.netlify.com/commands/env). The only change is that we rely on `--key` and `--value` instead of positional arguments to make it more descriptive. Open for alternatives though.

philnash:
In Heroku the `heroku config` commands set env variables. They are mostly the same as here (get, set, and unset all exist). Two things that might be different:

Calling `heroku config` lists the environment variables, saving the need for a `env:list` command. Though `list` commands do mostly fit with the CLIs existing patterns.

When calling `heroku config:set` you pass the environment variables as if they were environment variables. For example:

```
heroku config:set NAME=value
```

I like being more explicit about which the key and variable is, this is just a shorter hand version than using `--key` and `--value`.

One final thing, you have the commands setup as `twilio serverless:env get` but wouldn't this be `twilio serverless:env:get` along with the rest of the CLI?

dkundel:
Yeah I think making `list` an explicit option is nicer. As for `NAME=value` I think the only thing that I think makes my version preferred is combining it with existing environment variables. For example:

```
twilio serverless:env set --key TWILIO_API_KEY --value $TWILIO_API_KEY
```

As opposed to:

```
twilio serverless:env set TWILIO_API_KEY=$TWILIO_API_KEY
```

[...]

PR #290, opened 2021-06-25
feat: add new env commands

dkundel:
This change introduces a new set of commands around retrieving and modifying environment variables
from already deployed environments.

re #249

**Contributing to Twilio**


packages/plugin-serverless/src/commands/serverless/env/get.js
```diff
@@ -0,0 +1,41 @@
...
+  getRegionAndEdge,
+} = require('../../../utils');
+
+const { flags, aliasMap } = convertYargsOptionsToOclifFlags(cliInfo.options);
+
+class EnvGet extends TwilioClientCommand {
+  async run() {
+    await super.run();
+
+    let { flags, args } = this.parse(EnvGet);
```
philnash: You pointed out how I was using `this.args` and `this.flags` in my last PR. It turns out that is because the [`BaseCommand` does the arg/flag parsing already](https://github.com/twilio/twilio-cli-core/blob/main/src/base-commands/base-command.js#L35-L37). So you don't need to do it again and can just use `this.flags` and `this.args` too.

packages/plugin-serverless/src/commands/serverless/env/get.js
```diff
@@ -0,0 +1,41 @@
...
+const {
+  convertYargsOptionsToOclifFlags,
+  normalizeFlags,
+  createExternalCliOptions,
+  getRegionAndEdge,
+} = require('../../../utils');
+
+const { flags, aliasMap } = convertYargsOptionsToOclifFlags(cliInfo.options);
+
+class EnvGet extends TwilioClientCommand {
```
philnash: There is a lot of copy and paste! Any merit in creating a `TwilioRunCommand` that normalizes the flags, gets the region and edge, turns them all into `opts` and `externalOpts` and passes off to the `handler`? Or even just a factory function for this?

packages/serverless-api/src/client.ts
```diff
@@ -150,6 +161,125 @@ export class TwilioServerlessApiClient extends events.EventEmitter {
...
+        serviceSid,
+        this
+      );
+      environmentSid = environmentResource.sid;
+    } else {
+      environmentSid = config.environment;
+    }
+
+    return { serviceSid, environmentSid };
+  }
```
philnash, line 210: I know this is a draft PR, but this function is called `setEnvironmentVariables` but it only seems to collect a `serviceSid` and `environmentSid` and return them. Think it's missing the bit where it sets the vars! The `get` and `remove` functions do carry out their purpose, so I thought I'd call this one out.

philnash (changes requested):
Realise this is a draft, but I had a couple of suggestions, so I added them.

packages/serverless-api/src/client.ts
dkundel: Ha yeah it's not implemented yet 😄

packages/plugin-serverless/src/commands/serverless/env/get.js
dkundel: Yeah it's a good point! I'll do that

dkundel:
@philnash I implemented the class factory you suggested. I only implemented it for the new commands but I created #294 and #295 to implement it in the other commands and further remove the amount of copy paste code inside twilio-run.

packages/twilio-run/src/cli.ts
```diff
@@ -16,5 +17,14 @@ export async function run(rawArgs: string[]) {
     .command(ListCommand)
     .command(ActivateCommand)
     .command(LogsCommand)
+    .command(
+      'env',
+      'Retrieve and modify the environment variables for your deployment',
+      (yargs) => {
+        yargs.command(EnvCommands.GetCommand);
+        yargs.command(EnvCommands.ListCommand);
+        yargs.command(EnvCommands.UnsetCommand);
```
philnash, line 26: Are you missing `SetCommand` and `ImportCommand` here?

philnash:
That feels like a beautiful reduction in copy/paste! Everything else looks good.

philnash (approved):

philnash:
Also, makes sense to reuse this code once this has been merged to reduce the copy/paste elsewhere. Nice work!

PR #292, opened 2021-06-30
refactor(serverless-api): update user-agent string for better debugging

dkundel:
This PR standardizes the format of our User-Agent string in `@twilio-labs/serverless-api` and provide the ability to pass additional information from upstream packages to be included. It follows the agreed pattern of:

```
<core-api-lib>/<core-api-lib-version> (<os-name> <os-arch>) <extensions>
```

Where `<extensions>` are being added space separated following:

```
<library>/<library-version>
```

or

```
<tag>
```

In this case the `TwilioServerlessApiClient` offers to pass in `userAgentExtensions` as an array to specify upstream libraries who use it to add their library name and version.

**Contributing to Twilio**

### DIFF
--- a/packages/plugin-assets/src/init.js
+++ b/packages/plugin-assets/src/init.js
@@ -9,7 +9,7 @@ const {
 const { TwilioCliError } = require('@twilio/cli-core').services.error;
 
 const { couldNotGetEnvironment } = require('./errorMessages');
-const pkgJson = require('../package.json');
+const { getTwilioClient } = require('./client');
 
 async function createServiceAndEnvironment(client, serviceName) {
   const serviceSid = await createService(serviceName, client);
@@ -37,11 +37,7 @@ async function init({
   serviceName,
 }) {
   logger.debug('Loading config');
-  const client = new TwilioServerlessApiClient({
-    username: apiKey,
-    password: apiSecret,
-    userAgentExtensions: [`@twilio-labs/plugin-assets/${pkgJson.version}`],
-  });
+  const client = getTwilioClient(apiKey, apiSecret);
   const config = await pluginConfig.getConfig();
   if (
     config[accountSid] &&
--- a/packages/plugin-assets/src/init.js
+++ b/packages/plugin-assets/src/init.js
@@ -1,4 +1,3 @@
-const { TwilioServerlessApiClient } = require('@twilio-labs/serverless-api');
 const {
   createService,
 } = require('@twilio-labs/serverless-api/dist/api/services');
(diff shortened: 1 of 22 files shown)


packages/plugin-assets/src/init.js
```diff
@@ -39,6 +40,7 @@ async function init({
   const client = new TwilioServerlessApiClient({
     username: apiKey,
     password: apiSecret,
+    userAgentExtensions: [`@twilio-labs/plugin-assets/${pkgJson.version}`],
```
philnash: Suggestion, create `plugin-assets/src/client.js` with this:

```js
const { TwilioServerlessApiClient } = require('@twilio-labs/serverless-api');
const pkgJson = require("../package.json");

module.exports = function getTwilioServerlessApiClient(apiKey, apiSecret) {
  return new TwilioServerlessApiClient({
    username: apiKey,
    password: apiSecret,
    userAgentExtensions: [`@twilio-labs/plugin-assets/${pkgJson.version}`],
  });
}

```

Then in each of init.js, list.js and upload.js we can just require and use that function to instantiate a client.

This makes it easier to remember to create the client correctly as more commands are inevitably added in the future.

packages/serverless-api/src/client.ts
```diff
@@ -685,6 +698,10 @@ export class TwilioServerlessApiClient extends events.EventEmitter {
       statusCodes: [429],
       errorCodes: [],
     };
+    options.headers = {
+      ...options.headers,
+      'User-Agent': getUserAgent(),
+    };
```
philnash: Why are you adding the `User-Agent` to `this.client`'s headers again here?

packages/serverless-api/src/client.ts
```diff
@@ -87,7 +88,7 @@ export function createGotClient(config: ClientConfig): GotClient {
     username: username,
     password: password,
     headers: {
-      'User-Agent': 'twilio-serverless-api',
+      'User-Agent': getUserAgent(),
```
philnash: I think it would be better to pass `config.userAgentExtensions` to `getUserAgent` here, rather than using global scope to store a list of them. It would make `getUserAgent` easier to test and reason about (though I'm not going to call it "pure" given it reaches out to the OS for details).

philnash (changes requested):
Mostly looks good, added some ideas about simplifying a couple of bits.

dkundel:
I want to add at least one `nock` test specifically for the user-agent string before merging this but it's ready for review.

philnash (approved):
Looks good to me now!

PR #293, opened 2021-07-01
feat(runtime-handler): handle incoming headers and cookies

philnash:
This sets up the runtime handler to populate a `request` property on the event object with the `headers` and `cookies` that are sent to the endpoint.

**Contributing to Twilio**


philnash:
Still to come, in another PR, set cookies on the `Response` object.

packages/runtime-handler/src/dev-runtime/route.ts
```diff
@@ -35,8 +39,51 @@ const RUNNER_PATH =
...
 let twilio: TwilioPackage;
 
+type Headers = {
+  [key: string]: string | string[];
+};
+type Cookies = {
+  [key: string]: string;
+};
+
+export function constructHeaders(rawHeaders?: string[]): Headers {
```
dkundel, line 49: interesting choice to use `rawHeaders`. I think that makes sense especially because the regular headers has multi valued headers concatenated. However, I think there is some additional logic we need here because some headers cannot be multi-valued. 

Specifically you can see the list and how it's handled here: https://nodejs.org/api/http.html#http_message_headers

We should double check if this is the behavior that happens in production or not.

dkundel (changes requested):

packages/runtime-handler/src/dev-runtime/route.ts
dkundel: I know this is specifically for the parsed headers but I'm not sure which logic is being performed in our production environment.

philnash: Yeah, I used `rawHeaders` to avoid having to parse the multivalued headers back out of the `headers` object.

Agree that we need to confirm how these non-multivalued headers are handled in production to match them.

dkundel (approved):

PR #307, opened 2021-07-12
feat(logs): log polling backs off when it doesn't receive results

### LINKED ISSUE #251: Implement back off for logs streaming
Currently the logs command with `--tail` option keeps making requests to the logs endpoint at the rate of 1 request per second. There is also the ability to choose the period, though that is not currently exposed in CLI flags.

To lower the pressure on the logs endpoint, we propose to back off requests if no new logs are returned for a period of time.

Initial ideas were to reduce the period if no new logs are returned for 10 periods. After that we can double the period and wait 10 more periods before backing off again.

Once a new log is discovered and logged, the period should return to the initial period.

Each time we back off we should log the new period so that developers can see what is happening and either stop streaming, or restart the command because they expect to see new logs again.

philnash:
Fixes #251

**Contributing to Twilio**

### DIFF
--- a/packages/serverless-api/src/api/logs.ts
+++ b/packages/serverless-api/src/api/logs.ts
@@ -5,11 +5,39 @@ import { LogApiResource, LogList, Sid, LogFilters } from '../types';
 import { TwilioServerlessApiClient } from '../client';
 import { getPaginatedResource } from './utils/pagination';
 import { ClientApiError } from '../utils/error';
+import { OptionsOfJSONResponseBody } from 'got';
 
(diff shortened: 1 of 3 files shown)


dkundel:
What happens if we back off and then lots of logs (more than one page) get generated between the next poll. It doesn't look like we would currently fetch those again is that correct?

philnash:
In the serverless-api methods we don't expose the pagination, so I had to write a new function for this particular method (to avoid changing results and making a breaking change).

Might it be worth changing the list methods for the resources here to expose pagination so that external modules can choose one or multiple pages when they use this?

Issue #304, opened 2021-07-08
Local Development equivalent of SERVICE_SID and ENVIRONMENT_SID


dkundel:
The Functions team is introducing `SERVICE_SID` and `ENVIRONMENT_SID` to the `context` object. Since these are deployment related we need to decide what the behavior should be for local development.

Today we have two other deployment related variables `DOMAIN_NAME` and `PATH` but both of these can be set to equivalent local versions with `localhost:3000` (or an ngrok domain if tunneled) and the actual URL path. For the new two values there are technically no real equivalents.

## Options for Local Development

**PLEASE VOTE BY USING THE CORRESPONDING EMOJI**. Or provide more options in the comments below.

### Option 1 (❤️): `undefined`

Since `SERVICE_SID` and `ENVIRONMENT_SID` are tied to a deployment, my first thought lands on them not being available at all in local development.

Since the most common use case for them is to be used to dynamically set an environment variable the logic would look something like this:

```js
if (context.SERVICE_SID && context.ENVIRONMENT_SID) {
	context.getTwilioClient()
		.serverless.services(context.SERVICE_SID)
		.environments(context.ENVIRONMENT_SID)
		.variables.create({ key: 'DEMO', value: 'SOMETHING' });
}
```

### Option 2 (🚀): `undefined` & Warning

In this situation we'd also set the values to `undefined` but we'd implement them as getters using `Object.defineProperty` where we also emit a warning to the customer during local development. Something along the lines of:

```
WARN During local development SERVICE_SID will always be undefined. Make sure test for this possibility when using the value.
```

In theory this can both educate people but also confuse people. We might want to create a dedicated docs article in this case that explains the situation.

The code above would still be valid in this situation.

### Option 3 (😄): Fake value

In this case we'd set the value to the same pattern as the SIDs would be in production but with placeholders. For example `ZS11111111111111111111111111111111`. 

The main problem I foresee is that it makes it harder to gate code during local development and it would actually not provide any value other than knowing which format the future values would be.

The only befit here would be that the TypeScript types would be always `string` as opposed to `string | undefined`.

### Option 4 (🎉): "Magic"

In this situation we'd go with the same direction as Option 3 except that we set the value to some arbitrary magic value and then combine that with intercepting certain HTTP requests from Functions that go to the Serverless API. Meaning if a customer would write code such as:

```js
context.getTwilioClient()
	.serverless
	.services(context.SERVICE_SID)
[...]

PR #317, opened 2021-07-14
fix(runtime-handler): add warning for optional context vars

dkundel:
I'm going with Option 2 from #304 here as it got the most upvotes. We'll release it as `@twilio/runtime-handler/1.1.2` but in theory Option 2 is compatible with any version of the runtime-handler.

The warning looks the following:
Screen Shot 2021-07-13 at 17 53 21

re #304

**Contributing to Twilio**


philnash (approved):
Who doesn't love a regex over a a stack trace from an error that doesn't actually get thrown!

PR #319, opened 2021-07-16
feat(create-twilio-function): add deployinfo file to gitignore

### LINKED ISSUE #318: Add ..twiliodeployinfo to .gitignore
Hi,

I've done some public deploys and noticed that the .twiliodeployinfo contains my Account SID. Not a train smash, but it is now public if I use a public repo. I have manually added .twiliodeployinfo to .gitignore, but just wondering if we could build it into the default .gitignore file when doing a serverless:init?

Would be a nice to have

dkundel:
This change adds .twiliodeployinfo to the default .gitignore

fix #318

**Contributing to Twilio**

### DIFF
--- a/packages/create-twilio-function/src/create-twilio-function/create-gitignore.js
+++ b/packages/create-twilio-function/src/create-twilio-function/create-gitignore.js
@@ -1,20 +1,25 @@
-const fs = require('fs');
+const fs = require('fs').promises;
+const { createWriteStream } = require('fs');
 const path = require('path');
 const { promisify } = require('util');
 
 const writeGitignore = promisify(require('gitignore').writeFile);
 
-const open = promisify(fs.open);
+const ADDITIONAL_CONTENT = `
+# Twilio Serverless
+.twiliodeployinfo
+`;
 
-function createGitignore(dirPath) {
+async function createGitignore(dirPath) {
   const fullPath = path.join(dirPath, '.gitignore');
-  return open(fullPath, 'wx').then((fd) => {
-    const stream = fs.createWriteStream(null, { fd });
-    return writeGitignore({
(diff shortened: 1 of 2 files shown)


packages/create-twilio-function/src/create-twilio-function/create-gitignore.js
```diff
@@ -1,20 +1,25 @@
...
-      file: stream,
-    });
+  const fd = await fs.open(fullPath, 'wx');
+  const stream = createWriteStream(fullPath, { fd });
+  await writeGitignore({
+    type: 'Node',
+    file: stream,
   });
+  await fd.close();
+  await fs.appendFile(fullPath, ADDITIONAL_CONTENT, 'utf8');
```
philnash, line 22: Rather than close the file descriptor and open it again, you could do:

```suggestion
await fd.appendFile(ADDITIONAL_CONTENT, 'utf8');
await fd.close();
```

philnash (changes requested):

packages/create-twilio-function/src/create-twilio-function/create-gitignore.js
dkundel: That unfortunately didn't work. I've tried that for a while and gave up because the other thing worked.

philnash: Hmm, you're right. That's because the `gitignore` package closes the stream itself. And perhaps the open file descriptor. When I tried this suggested method, the tests passed, but doing it for real did not work (event weirder).

I'm having a go at rewriting this because the confusion between promises and streams (which don't support promises in the version of Node we are using) are making this harder to deal with. PR incoming.

PR #330, opened 2021-07-22
chore(plugin-serverless): remove globby devDependency

philnash:
It does not seem to be in use, so removing from plugin-serverless.

@dkundel it looks like you added this dependency, but I can't see it in use at all. Am I missing something or can we safely remove it?

**Contributing to Twilio**


philnash:
Should supersede #327

PR #345, opened 2021-09-13
Json output

philnash:
Adds an `--output-format` (or `-o`) flag to the `deploy`, `list-templates`, `list` and `promote` commands. Commands that already supported this flag include `logs` and `env:list`. Commands that do not support an output format currently include `new` and `start`.

This supersedes #143.

**Contributing to Twilio**

### DIFF
--- a/packages/runtime-handler/src/dev-runtime/server.ts
+++ b/packages/runtime-handler/src/dev-runtime/server.ts
@@ -1,6 +1,6 @@
 import { ServerlessFunctionSignature } from '@twilio-labs/serverless-runtime-types/types';
 import bodyParser from 'body-parser';
-import EventEmitter from 'events';
+import { EventEmitter } from 'events';
 import express, {
   Express,
   NextFunction,
--- a/packages/twilio-run/src/commands/shared.ts
+++ b/packages/twilio-run/src/commands/shared.ts
@@ -11,3 +11,5 @@ export type ExternalCliOptions = {
     version: string;
   };
 };
+
+export type OutputFormat = string | undefined;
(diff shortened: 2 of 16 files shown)


packages/runtime-handler/src/dev-runtime/server.ts
```diff
@@ -1,6 +1,6 @@
 import { ServerlessFunctionSignature } from '@twilio-labs/serverless-runtime-types/types';
 import bodyParser from 'body-parser';
-import EventEmitter from 'events';
```
dkundel, line 3: What's the purpose of this change?

packages/twilio-run/src/commands/shared.ts
```diff
@@ -11,3 +11,5 @@ export type ExternalCliOptions = {
     version: string;
   };
 };
+
+export type OutputFormat = string | undefined;
```
dkundel, line 15: should this be:
```suggestion
export type OutputFormat = 'json' | undefined;
```

packages/twilio-run/src/config/deploy.ts
```diff
@@ -112,6 +114,7 @@ export async function getConfigFromFlags(
   }
 
   const { region, edge, runtime } = flags;
+  const outputFormat = flags.outputFormat || externalCliOptions?.outputFormat;
```
dkundel, line 117: Do we still need the concept of `externalCliOptions.outputFormat` if we support this as a flag?

dkundel (commented):
Left a couple of comments. Looks good overall

packages/twilio-run/src/commands/shared.ts
philnash: That clashes with the `externalCliOptions` version of `outputFormat` which has it set to `string | undefined`.

PR #359, opened 2021-09-27
Update dependencies

philnash:
This covers all the recent Snyk updates, ensuring the tests pass and the TS apps compile.

Snyk has suggested some updates like wrap-ansi@8 or ora@6 which are ESM only modules, which don't currently work with our setup. I will raise an issue to start a discussion about what to do with this sort of module.

**Contributing to Twilio**

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   "devDependencies": {
     "@commitlint/cli": "^9.1.2",
     "@commitlint/config-conventional": "^10.0.0",
-    "@types/jest": "^24.0.15",
+    "@types/jest": "^26.0.24",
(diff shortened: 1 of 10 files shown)


dkundel:
There is a variety of major version changes in here. Have you checked the respective Changelogs to make sure we don't have any side effects based on these changes?

philnash:
Yeah, major version changes are dropping support for Node versions below 12. Since 12 is the floor for us already, this isn't an issue. I also ran some manual tests to make sure things still worked as expected.

dkundel (approved):

philnash:
Will merging this close all the Snyk PRs?

PR #369, opened 2021-10-06
feat: Merge type files into bundle

mbeldartwilio:
Currently, `Monaco` does not support importing multiple files.

This PR adds a script which merges `index.d.ts` and `types.d.ts` into `bundle.d.ts`.
Script is run at `npm prepare` stage.
**Contributing to Twilio**

### DIFF
--- a/packages/serverless-runtime-types/package.json
+++ b/packages/serverless-runtime-types/package.json
@@ -6,7 +6,8 @@
   "types": "index.d.ts",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
-    "build": "sh bundler.sh"
+    "build": "sh bundler.sh",
+    "prepare": "npm run-script build"
   },
   "keywords": [
     "twilio",
--- a/packages/serverless-runtime-types/package.json
+++ b/packages/serverless-runtime-types/package.json
@@ -6,7 +6,7 @@
   "types": "index.d.ts",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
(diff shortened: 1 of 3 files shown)


dkundel (approved):

packages/serverless-runtime-types/package.json
```diff
@@ -5,7 +5,9 @@
   "main": "index.js",
   "types": "index.d.ts",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "build": "./bundler.sh",
+    "prepare": "npm run-script build"
```
dkundel: Can you change this to `prepack` instead? And `npm run` should be sufficient.

packages/serverless-runtime-types/package.json
```diff
@@ -5,7 +5,9 @@
   "main": "index.js",
   "types": "index.d.ts",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "build": "./bundler.sh",
```
dkundel: It looks like this is failing the CI platform on Windows because it can't execute the shell script. There might be a way to solve this elegantly but an alternative could be to change the `bundler.sh` script into a few lines of Node.js script using `fs.readFileSync()` and `fs.writeFileSync()`.

dkundel (changes requested):

dkundel (approved):

PR #370, opened 2021-10-08
fix(twilio-run): support exact dependency ranges in templates

### LINKED ISSUE #365: Exact dependencies in templates are not correctly installed
It seems like when we trigger pkg-install the dependencies are automatically installed in version ranges even if an exact version is given.

Example:

```
twilio serverless:init example --template=verified-broadcast
```

Will result in a `@twilio/runtime-handler` of `^1.2.0-rc.3` even though it should be an exact version the same way as in the template.

This might be solved if we would pass in the exact flag to pkg-install but this will require more research to validate that it will preserve ranges.

dkundel:
npm install does not preserve the passed dependency ranges resulting in templates that have exact
dependency versions specified but get added with a semver range to the package.json. This change
installs exact dependencies separately with the --save-exact flag.

fix #365

**Contributing to Twilio**

### DIFF
--- a/packages/twilio-run/__tests__/templating/filesystem.test.ts
+++ b/packages/twilio-run/__tests__/templating/filesystem.test.ts
@@ -376,6 +376,129 @@ test('installation with a dependency file', async () => {
   );
 });
 
+test('installation with a dependency file with exact dependencies', async () => {
+  // The typing of `got` is not exactly correct for this - it expects a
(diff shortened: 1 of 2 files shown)


philnash (approved):
Looks good to me. Took me a while to realise that `semver.parse` would return `null` for a string like "^3.0.0" but once I got that it makes sense.

Issue #371, opened 2021-10-14
Bare path for /assets/index.html


dkundel:
For legacy support reasons if you upload an asset as `/assets/index.html` any request to `/` will automatically serve that file. This isn't currently supported in the local development experience.

PR #372, opened 2021-10-14
Fix for #371

dkundel:
This PR implements the existing logic for how `/` will fallback to `/assets/index.html` in production.

**Contributing to Twilio**

### DIFF
--- a/packages/runtime-handler/src/dev-runtime/server.ts
+++ b/packages/runtime-handler/src/dev-runtime/server.ts
@@ -141,12 +141,15 @@ export class LocalDevelopmentServer extends EventEmitter {
     app.all(
       '/*',
       (req: ExpressRequest, res: ExpressResponse, next: NextFunction) => {
-        if (!this.routeMap.has(req.path)) {
-          res.status(404).send('Could not find request resource');
-          return;
+        let routeInfo = this.routeMap.get(req.path);
+
+        if (!routeInfo && req.path === '/') {
+          log('Falling back to /assets/index.html');
+          // In production we automatically fall back to the contents of /assets/index.html
+          routeInfo = this.routeMap.get('/assets/index.html');
         }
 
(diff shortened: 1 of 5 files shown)


packages/runtime-handler/src/dev-runtime/server.ts
```diff
@@ -141,12 +141,15 @@ export class LocalDevelopmentServer extends EventEmitter {
     app.all(
       '/*',
       (req: ExpressRequest, res: ExpressResponse, next: NextFunction) => {
-        if (!this.routeMap.has(req.path)) {
-          res.status(404).send('Could not find request resource');
-          return;
+        let routeInfo = this.routeMap.get(req.path);
```
stern-shawn, line 144: I cry every time I see a `const` convert to a `let` 😢

packages/runtime-handler/src/dev-runtime/server.ts
```diff
@@ -141,12 +141,15 @@ export class LocalDevelopmentServer extends EventEmitter {
...
+
+        if (!routeInfo && req.path === '/') {
+          log('Falling back to /assets/index.html');
+          // In production we automatically fall back to the contents of /assets/index.html
+          routeInfo = this.routeMap.get('/assets/index.html');
         }
 
-        if (req.method === 'OPTIONS') {
+        if (req.method === 'OPTIONS' && routeInfo) {
```
stern-shawn, line 152: Is this check for non-falsy `routeInfo` important? Ex. we only want to 204 OPTIONS requests to defined paths?

stern-shawn (approved):
Looks good to me with one question on the `OPTIONS` case 🚢 🇮🇹

packages/runtime-handler/src/dev-runtime/server.ts
dkundel: Yeah because that's how the logic was before.

PR #397, opened 2022-06-01
[Security ] twilio 3.29.* to 3.71.*

shadywallas:
Twilio 3.29 package includes an insecure version of `moment.js` 
https://github.com/advisories/GHSA-8hfj-j24r-96c4

**Contributing to Twilio**

### DIFF
--- a/packages/runtime-handler/package.json
+++ b/packages/runtime-handler/package.json
@@ -73,7 +73,7 @@
     "nocache": "^2.1.0",
     "normalize.css": "^8.0.1",
     "serialize-error": "^7.0.1",
-    "twilio": "3.29.2"
+    "twilio": "^3.71.2"
   },
   "gitHead": "6db273648ed19474f4125042556b10c051529912"
 }


philnash:
Thank you for taking a look at this. However this isn't a patch we can accept for now. The runtime-handler is designed to mimic the behaviour of Twilio Functions, which currently defaults to Twilio version 3.29.2. Changing that default could have consequences for applications that rely on that version and would need a plan to make that upgrade. So, this public runtime-handler will need to continue to follow that behaviour.

This should not be a problem for a couple of reasons.

Firstly, if you are building an application using Twilio Functions you can upgrade the version of the Twilio dependency yourself in your project (either in the dependencies section of the Twilio Functions UI, or if you are using the Serverless Toolkit, by updating the version in your package.json).

Secondly, the vulnerability seems to lie in the code that sets a locale for a date in moment, which the Twilio package didn't use. You can see where [moment was removed from the package here](https://github.com/twilio/twilio-node/pull/515/files) and the usage is minimal.

Thank you again for looking into this though. We always appreciate contributions, even when we can't currently accept them. Do let me know if you'd like to help out anywhere else, I'd be happy to help.

PR #410, opened 2022-08-02
chore(deps): upgrades typescript to 4.2.4

philnash:
Fixes one type issue that came up in the upgrade. The project builds now.

4.2.4 is not the latest version of TS for now. Reports around the internet showed that upgrading to TS 4.2 fixed the Prettier issues we've been seeing. Upgrading all the way to 4.7.x caused a bunch more type issues. We should plan an upgrade, in order to keep up, but there might be further effects.

**Contributing to Twilio**

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "prettier": "^2.2.1",
     "rimraf": "^3.0.2",
     "ts-jest": "^26.5.0",
-    "typescript": "^3.9.7"
+    "typescript": "^4.2.4"
   },
   "lint-staged": {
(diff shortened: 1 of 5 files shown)


philnash:
ARGH. Those failures do not happen on my machine. (╯°□°)╯︵ ┻━┻

I'll remove all my node_modules and try again tomorrow.

dkundel:
I found an alternative fix for now. We should still upgrade TypeScript but it's more of a pain. #412 introduces a temporary fix

PR #412, opened 2022-08-05
Fix vulnerabilities in packages

dkundel:
- Fixes builds by overriding the `@types/prettier` for all impacted packages
- Updates all vulnerable dependencies
- Updates TypeScript for `@twilio-labs/serverless-api` so it can handle the `file-type` package
- Removes Node.js 12 from package generation
- Turns off the unit tests that were testing `file-type` until we can figure out what is the issue with Jest + TypeScript + ESM/CJS
- Turns off Typedoc until we figure out how to fix the configuration with the new major version

**Contributing to Twilio**


dkundel:
I'm aware that the deployment currently doesn't work since I removed the generation of the Typedoc from the PR. I'll open a separate PR next week to re-enable them.

packages/serverless-api/tsconfig.json
```diff
@@ -2,7 +2,13 @@
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
     "declaration": true,
-    "outDir": "./dist" /* Redirect output structure to the directory. */
+    "outDir": "./dist" /* Redirect output structure to the directory. */,
```
bnb, line 5: Does TS allow comments in JSON?

packages/twilio-run/package.json
```diff
@@ -53,14 +53,15 @@
     "cosmiconfig": "^7.0.0",
     "debug": "^3.1.0",
     "dotenv": "^6.2.0",
+    "enquirer": "^2.3.6",
```
bnb, line 56: I see this is added but nothing in this PR uses it?

packages/serverless-api/package.json
```diff
@@ -10,7 +10,7 @@
     "build": "tsc",
     "watch": "tsc --watch",
     "build:noemit": "tsc --noEmit",
-    "docs": "typedoc --options typedoc.json",
+    "docs": "echo 'fix typedoc --options typedoc.json'; exit 1;",
```
bnb, line 13: Is it intended that `npm run docs` will now log a command and exit with a failure (non-zero)?

bnb (approved):
outside of TS changes, YAML and JSON changes look as I'd expect them to. I've left a few comments above but they're not blocking - my assumption is there's reasons for them that I'm just not aware of.

packages/twilio-run/package.json
bnb: Ah, I see your note about the typedoc. I'll resolve this since you addressed it already :)

packages/serverless-api/package.json
bnb: Ah, I see your note about the typedoc. I'll resolve this since you addressed it already :)

packages/serverless-api/tsconfig.json
dkundel: Yep :)

packages/twilio-run/package.json
dkundel: It's a peer dependency of listr2 that was giving issues if not installed separately

packages/serverless-api/package.json
dkundel: jep so I don't forget about it :D

packages/runtime-handler/package.json
```diff
@@ -73,7 +73,10 @@
     "nocache": "^2.1.0",
     "normalize.css": "^8.0.1",
     "serialize-error": "^7.0.1",
-    "twilio": "3.29.2"
+    "twilio": "3.80.0"
```
philnash, line 76: Late comment here, but shouldn't this match the default in Functions? And then we have the method that requires the project's Twilio dependency and falls back to this one? @dkundel

PR #427, opened 2022-10-06
chore: override broken @types package to avoid TS bump

dkundel:
Overriding a package that's incompatible with TS 3.9 (what we are currently using) as it's breaking the CI system. We should upgrade to the latest TypeScript major and pin @types packages respectively but that's a bigger effort.

**Contributing to Twilio**

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,8 @@
     "typescript": "^3.9.7"
   },
   "overrides": {
-    "@types/prettier": "2.6.0"
+    "@types/prettier": "2.6.0",
+    "@types/express-serve-static-core": "ts3.9"
   },
   "lint-staged": {
     "*.{js,jsx,ts,tsx}": [
--- a/packages/runtime-handler/package.json
+++ b/packages/runtime-handler/package.json
@@ -76,7 +76,8 @@
     "twilio": "3.80.0"
   },
   "overrides": {
-    "@types/prettier": "2.6.0"
+    "@types/prettier": "2.6.0",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -38,7 +38,8 @@
     "typescript": "^3.9.7"
   },
   "overrides": {
-    "@types/prettier": "2.6.0"
+    "@types/prettier": "2.6.0",
+    "@types/express-serve-static-core": "ts3.9"
```
bnb, line 42: Are these types needed here? It does not seem like this specific package.json has a dependency on Express. Perhaps this is just a part of how Lerna works that I'm not familiar with?

bnb (commented):
Are you okay with the override being relative? You're using a tag rather than a version so the version could change out from under you the future.

If so, this LGTM. If not, might want to pin to a specific version ([4.17.28](https://www.npmjs.com/package/@types/express-serve-static-core/v/4.17.28) is what the tag currently points to).

package.json
dkundel: I don't think it's strictly necessary here. Npm overrides gives me a headache at times and wanted to be extra sure that the wrong version can't be anywhere in the path especially since we run TypeScript typically from the root. There's definitely some overall dependency cleanup we can do.

dkundel:
I'm trying to see if tags actually work for this pretty reliably so that in the future we can tag more @types dependencies with the tag rather than version range. If it doesn't work I'll revert to the fixed version.

PR #433, opened 2022-10-28
feat(packages/twilio-run): regionalize toolkit config and api

mbeldartwilio:
**Contributing to Twilio**

### DIFF
--- a/packages/serverless-api/src/api/utils/api-client.ts
+++ b/packages/serverless-api/src/api/utils/api-client.ts
@@ -1,5 +1,13 @@
 import { ClientConfig } from '../../types';
 
+const regionEdgeMap: { [index: string]: string } = {
+  us1: 'ashburn',
+  au1: 'sydney',
+  ie1: 'dublin',
+  'stage-us1': 'ashburn',
+  'stage-au1': 'sydney',
+};
+
 export function getApiUrl(
   config: ClientConfig,
   product = 'serverless',
@@ -7,8 +15,15 @@ export function getApiUrl(
 ): string {
   const configEdge = config.edge || process.env.TWILIO_EDGE;
   const configRegion = config.region || process.env.TWILIO_REGION;
+  const region = configRegion ? `${configRegion}.` : '';
+
+  if (!configEdge && configRegion) {
+    const defaultEdge = regionEdgeMap[configRegion]
+      ? `${regionEdgeMap[configRegion]}.`
+      : '';
+    return `https://${product}.${defaultEdge}${region}twilio.com/${apiVersion}`;
+  }
 
   const edge = configEdge ? `${configEdge}.` : '';
-  const region = configRegion ? `${configRegion}.` : '';
   return `https://${product}.${edge}${region}twilio.com/${apiVersion}`;
 }
--- a/packages/twilio-run/__tests__/config/global.test.ts
+++ b/packages/twilio-run/__tests__/config/global.test.ts
@@ -145,4 +145,69 @@ describe('readSpecializedConfig', () => {
       env: '.env.stage',
     });
   });
+
+  test('account + region config override', () => {
+    __setTestConfig({
+      serviceSid: 'ZS11112222111122221111222211112222',
+      env: '.env.example',
+      commands: {
+        deploy: {
+          functionsFolder: '/tmp/functions',
+        },
+      },
+      environments: {
+        prod: {
+          serviceSid: 'ZS11112222111122221111222211112223',
+          env: '.env.prod',
+        },
+      },
+      projects: {
+        'AC11112222111122221111222211114444:au1': {
+          serviceSid: 'ZS11112222111122221111222211114444',
+        },
+        'AC11112222111122221111222211114444:ie1': {
+          serviceSid: 'ZS11112222111122221111222211114445',
+        },
+        AC11112222111122221111222211114444: {
+          serviceSid: 'ZS11112222111122221111222211114446',
+        },
+      },
+    });
+
+    expect(
+      readSpecializedConfig('/tmp', '.twilioserverlessrc', 'deploy', {
+        environmentSuffix: 'prod',
+        username: 'AC11112222111122221111222211114444',
+        region: 'ie1',
+      })
+    ).toEqual({
+      serviceSid: 'ZS11112222111122221111222211114445',
+      functionsFolder: '/tmp/functions',
+      env: '.env.prod',
+    });
+
+    expect(
+      readSpecializedConfig('/tmp', '.twilioserverlessrc', 'deploy', {
+        environmentSuffix: 'prod',
+        username: 'AC11112222111122221111222211114444',
+        region: 'au1',
+      })
+    ).toEqual({
+      serviceSid: 'ZS11112222111122221111222211114444',
+      functionsFolder: '/tmp/functions',
+      env: '.env.prod',
+    });
+
+    expect(
+      readSpecializedConfig('/tmp', '.twilioserverlessrc', 'deploy', {
+        environmentSuffix: 'prod',
+        username: 'AC11112222111122221111222211114444',
+        region: 'us1',
+      })
+    ).toEqual({
+      serviceSid: 'ZS11112222111122221111222211114446',
+      functionsFolder: '/tmp/functions',
+      env: '.env.prod',
+    });
+  });
 });
--- a/packages/twilio-run/src/config/deploy.ts
+++ b/packages/twilio-run/src/config/deploy.ts
@@ -99,7 +99,8 @@ export async function getConfigFromFlags(
         ? flags.username
         : username.startsWith('AC')
         ? username
-        : externalCliOptions?.accountSid
+        : externalCliOptions?.accountSid,
+      flags.region
     ));
 
   const pkgJson = await readPackageJsonContent(flags);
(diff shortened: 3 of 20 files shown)


packages/twilio-run/src/serverless-api/utils.ts
```diff
@@ -35,6 +36,18 @@ export async function getFunctionServiceSid(
   if (username) {
     debug('Attempting to read serviceSid from a deployinfo file');
     const deployInfoCache = getDeployInfoCache(cwd);
+    if (
+      deployInfoCache &&
+      deployInfoCache[`${username}:${region}`] &&
+      deployInfoCache[username].serviceSid
```
dkundel:
```suggestion
      deployInfoCache[`${username}:${region}`].serviceSid
```
Otherwise this if statement would fail.

Can you also make sure to add a fallback if `deployInfoCache[username]` is available but `deployInfoCache[`${username}:${region}`]` is not available?

packages/twilio-run/src/utils/deployInfoCache.ts
```diff
@@ -71,9 +72,15 @@ export function updateDeployInfoCache(
     deployInfoCacheFileName
   );
 
+  if (username in currentDeployInfoCache) {
```
dkundel: Instead of `in` can use use `currentDeployInfoCache[username]` or even `currentDeployInfoCache.hasOwnProperty(username)`? 

The other thing to consider is that right now this would delete even if you don't deploy to that region. Let's say you deploy to `ie1` it would delete the old entry entirely. So I think we should only do that if `region === 'us1'`.

dkundel (changes requested):
Excited for this :) Left a couple of initial comments

packages/twilio-run/src/serverless-api/utils.ts
mbeldartwilio: For fallback, I have kept the logic as is [here](https://github.com/twilio-labs/serverless-toolkit/pull/433/files#diff-3a6a3698899b9890e5faaef4af544dcf18f4bd8f68a1d7edf335fbedbf81a2d5R51-R62)

Would that suffice?

dkundel: Ah yeah that should work. It was collapased in the code review tab

packages/serverless-api/src/api/utils/api-client.ts
```diff
@@ -1,14 +1,29 @@
...
   config: ClientConfig,
   product = 'serverless',
   apiVersion = 'v1'
 ): string {
   const configEdge = config.edge || process.env.TWILIO_EDGE;
   const configRegion = config.region || process.env.TWILIO_REGION;
+  const region = configRegion ? `${configRegion}.` : '';
+
+  if (!configEdge && configRegion) {
+    const defaultEdge = regionEdgeMap[configRegion]
```
vingiarrusso, line 21: It looks like this will silently ignore invalid regions and replace them with default.  Should we throw an error back to the user here and let them know the region they specified is not a valid region for Serverless API?

mbeldartwilio, line 21: I remember Dom mentioning, we should let toolkit fallback to the API and be dumb about this.

packages/twilio-run/src/serverless-api/utils.ts
```diff
@@ -35,6 +37,18 @@ export async function getFunctionServiceSid(
   if (username) {
     debug('Attempting to read serviceSid from a deployinfo file');
     const deployInfoCache = getDeployInfoCache(cwd);
+    if (
+      deployInfoCache &&
+      deployInfoCache[`${username}:${region}`] &&
+      deployInfoCache[`${username}:${region}`].serviceSid
```
vingiarrusso: can we use optional chaining here?  ```if (deployInfoCache[`${username}:${region}`]?.serviceSid) ...```

packages/serverless-api/src/api/utils/api-client.ts
vingiarrusso: ~~Hm.  Yes, but in this case, it's not exactly expected or desired behavior imo.  If i recall the conversation correctly, we discussed if the user specifies an edge we don't have, and then we try to use that edge in the URL, then they'd get a 404 back which isn't great or specific, but at least it would fail.  In this case, you specify an invalid region and end up with something deployed to US1, which is kind of unexpected.  @dkundel thoughts?~~

edit: misread what this did

vingiarrusso: oh wait, im totally misreading!  my fault.  That looks like what this will do.  Okay all good :)

packages/twilio-run/src/serverless-api/utils.ts
classactcollin: You should be able to since the node version is >14.

packages/twilio-run/src/config/global.ts
```diff
@@ -68,12 +69,20 @@ export function readSpecializedConfig<T extends CommandConfigurationNames>(
...
-    result = {
-      ...result,
-      ...projectsConfig[opts.username],
-    };
+    if (projectsConfig[`${opts.username}:${opts.region}`]) {
+      result = {
+        ...result,
+        ...projectsConfig[`${opts.username}:${opts.region}`],
+      };
+    } else if (opts.region === 'us1' || opts.region === undefined) {
```
vincianechen, line 80: do we need to put this first? not sure if we're allowing `username:us1`, but it seems like the `opts.region === 'us1'` case would fall into the if block, so we should remove it or swap the if and else if blocks.

packages/twilio-run/src/utils/deployInfoCache.ts
```diff
@@ -71,9 +72,15 @@ export function updateDeployInfoCache(
     deployInfoCacheFileName
   );
 
+  if (currentDeployInfoCache.hasOwnProperty(username) && region === 'us1') {
+    debug('Invalid format for deploy info key. Overriding with region us1');
```
vincianechen, line 76: i'm confused about this block, is the username always an account sid? if so, it seems like this would only be invalid if just the account sid key is in the cache, but the region is not `us1`? lmk if i'm misunderstanding though.

packages/twilio-run/__tests__/config/global.test.ts
```diff
@@ -145,4 +145,69 @@ describe('readSpecializedConfig', () => {
       env: '.env.stage',
     });
   });
+
+  test('account + region config override', () => {
+    __setTestConfig({
+      serviceSid: 'ZS11112222111122221111222211112222',
```
vincianechen, line 151: the sid convention is a little confusing here, is this service sid supposed to match up with any of the ones below? or does only the service sids in `projects` matter for this test?

packages/twilio-run/src/serverless-api/utils.ts
mbeldartwilio: Ideally, this project should support node12 and above (from readme)
I would like to know Dom's opinion as node12 is deprecated already.

dkundel: I'm okay with moving to Node 14 now. I think that's outdated in the README since twilio-run itself already only supports Node 14 since a few versions ago

packages/twilio-run/src/config/global.ts
mbeldartwilio: I think logic you mentioned would fail in following case:
`region=us1` and the config file with `account:us1`.

And yes, we will be allowing accountsid:us1

packages/twilio-run/src/utils/deployInfoCache.ts
mbeldartwilio: this is handling the case where user has passed `us1` in new build request but the existing config is still using 'username' and not 'username:us1'.

We wan't to remove 'username' as we are adding 'username:us1'. (see next few lines)

packages/twilio-run/__tests__/config/global.test.ts
mbeldartwilio: from what I understand, the ZS...22 is default servicesid. Other service sids under "envs" and "accounts" supposed to override the default one.

vingiarrusso (approved):

packages/twilio-run/src/serverless-api/utils.ts
```diff
@@ -49,20 +55,22 @@ export async function getFunctionServiceSid(
   }
 
   debug('Could not determine existing serviceSid');
+  debug(`${username}:${region}`);
   return undefined;
 }
 
 export async function saveLatestDeploymentData(
```
dkundel, line 62: It looks like you are missing in the `packages/twilio-run/src/commands/deploy.ts` file an update to the call of `saveLatestDeploymentData` that passes `config.region`.

packages/twilio-run/src/config/deploy.ts
```diff
@@ -74,6 +74,7 @@ export async function getConfigFromFlags(
       (externalCliOptions && externalCliOptions.accountSid) ||
       undefined,
     environmentSuffix: flags.environment,
+    region: flags.region,
```
dkundel, line 77: The calls to `getFunctionServiceSid` are missing the `region` passed in as the cache from the `twilioDeployInfo` file are not being read. Same applies in the other files.

dkundel (changes requested):
The `.twilioserverlessrc` behavior changes and the edge fallbacks work flawlessly. For the `.twiliodeplyinfo` changes it looks like the regions are not being passed for read and write all the way down. Can you change that? :)

dkundel (approved):

PR #446, opened 2022-12-09
fix(twilio-run): remove key/value check from env:import command

dkundel:
The `serverless:env:import` command wasn't working as expected because we had a check for `key` and `value` which aren't actually required. This appears to be a copy & paste mistake because those two flags aren't even used in the code.

**Contributing to Twilio**

### DIFF
--- a/packages/twilio-run/src/config/env/env-import.ts
+++ b/packages/twilio-run/src/config/env/env-import.ts
@@ -79,18 +79,6 @@ export async function getConfigFromFlags(
 
   let serviceName = await getServiceNameFromFlags(flags);
 
-  if (!flags.key) {
-    throw new Error(
-      'Missing --key argument. Please provide a key for your environment variable.'
-    );
-  }
-
-  if (!flags.value) {
-    throw new Error(
-      'Missing --value argument. Please provide a key for your environment variable.'
-    );
-  }
-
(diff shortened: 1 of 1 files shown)


packages/twilio-run/src/config/env/env-import.ts
```diff
@@ -79,18 +79,6 @@ export async function getConfigFromFlags(
...
 
-  if (!flags.key) {
-    throw new Error(
-      'Missing --key argument. Please provide a key for your environment variable.'
-    );
-  }
-
-  if (!flags.value) {
-    throw new Error(
-      'Missing --value argument. Please provide a key for your environment variable.'
```
gautam678, line 90: Should this be `value`?

gautam678 (approved):

packages/twilio-run/src/config/env/env-import.ts
dkundel: They are both not used so I'm just deleting the checks. They were copy & paste from the env-set.ts file which actually needs them.

gautam678: Oh nvm, I thought you were adding this 🙈

pthirumurthi (approved):

PR #450, opened 2022-12-30
Remove Lerna

dkundel:
**Work Included**:
- [x] Replace `lerna` for bootstrapping and task running in favor of `npm workspaces`
- [x] Upgrade TypeScript to 4.9.4
- [x] Upgrade Jest and ts-jest
- [x] Replace  `lerna` in favor of `changesets` for versioning
- [x] Setup release pipeline through GitHub Actions similar to github.com/twilio-labs/plugin-webhook

**Known Issues**:
- [x] ~The `requireFromProject.test.ts` files are currently partially disabled because `npm workspaces` hoists the `twilio` dependency to the root `node_modules` folder instead of having multiple instances. This results in the test no longer working. Virual mocks might be an alternative but needs validation.`~
- [ ] `skipLibCheck` is currently enabled in the `tsconfig.base.json` because `@types/inquirer` and `execa` were throwing issues --> Moved to #451 
- [ ] `useUnknownInCatchVariables` is currently disabled as it was a breaking change between TypeScript 3.x and 4.4 and requires code changes. --> Moved to #452 
- [x] ~Netlify build is currently failing.~

**Contributing to Twilio**


.github/workflows/nodejs.yml
```diff
@@ -19,10 +21,10 @@ jobs:
       fail-fast: false
       matrix:
         os: [macos-latest, windows-latest, ubuntu-latest]
-        node-version: [14, 16]
+        node-version: [14, 16, 18]
```
bnb: might want to use the relative notation here if you want automatic CI upgrades - totally understandable if not!

```suggestion
        node-version: [ lts/-2, lts/-1, lts/*]
```

.github/workflows/nodejs.yml
```diff
@@ -19,10 +21,10 @@ jobs:
...
 
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
```
bnb: v3 is most current if you're updating this anyway

.github/workflows/on-merge-main.yml
```diff
@@ -0,0 +1,42 @@
...
+    uses: twilio-labs/serverless-toolkit/.github/workflows/nodejs.yml@main
+
+  release:
+    name: Run release work
+    needs: tests
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v2
```
bnb: v3 is latest. Please feel free to resolve if you don't want to update to v3!

.github/workflows/nodejs.yml
```diff
@@ -19,10 +21,10 @@ jobs:
...
       - name: Use Node.js ${{ matrix.node-version }}
         uses: actions/setup-node@v1
```
bnb: If you do want to use relative versions, this would need to get updated to v3 as well.

.github/workflows/on-merge-main.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Setup Node.js 16
+        uses: actions/setup-node@v2
```
bnb: same v3 notes as previous :)

.github/workflows/on-merge-main.yml
```diff
@@ -0,0 +1,42 @@
...
+        with:
+          node-version: 16
+      - name: Update npm to 8.x
+        run: |
+          npm i -g npm@8
```
bnb, line 27: is there a reason here to use 8 instead of 9? 👀

bnb (commented):
some notes, mostly about version choices. None of these are blocking, LGTM otherwise :)

package.json
```diff
@@ -2,46 +2,45 @@
...
+    "version-packages": "npm run create-version && npm run version-package --workspaces --if-present",
+    "create-version": "changeset version",
+    "npm:publish": "changeset publish",
     "cm": "git-cz",
     "jest": "jest",
-    "build:noemit": "lerna run build:noemit",
-    "clean": "lerna run clean",
+    "build:noemit": "npm run build:noemit --workspaces --if-present",
+    "clean": "npm run clean --workspaces --if-present",
     "test": "run-s build:noemit jest",
```
bnb, line 15: one thing I realized I should/could advise on here: I'm not sure what `run-s` is but this might be able to be replaced with npm commands if `build:noemit jest` are scripts in each workspace. Specifically `npm run build:noemit --workspaces && npm run jest --workspaces`.

dkundel: `run-s` is an alias from `npm-run-all` and runs those two comments in serial. It's a holdover from the time when `&&` wouldn't always work on Windows. It could probably go away now because I think `npm` scripts make `&&` work even on Windows. So the equivalent here would be `npm run build:noemit && npm run jest`

.github/workflows/nodejs.yml
```diff
@@ -1,11 +1,13 @@
...
-      - main
-      - master
-      - 'features/*'
+  workflow_call:
+  # This gets run by on-merge-main.yml automatically
+  # push:
+  #   branches:
+  #     - main
+  #     - master
+  #     - 'features/*'
```
gautam678: Should this comment be removed?

docs/CONTRIBUTING.md
```diff
@@ -14,10 +14,10 @@ If you are planning to contribute something that does not have an open issue yet
 
 ## Requirements
 
-Make sure you have Node.js 12 or newer installed. Due to compatibility with Twilio
-Functions this project has to support at least Node.js 12.0.0.
+Make sure you have Node.js 14 or newer installed. Due to compatibility with Twilio
+Functions this project has to support at least Node.js 14.0.0.
```
gautam678, line 18: We could start with Node 16, since that will be used as default here. What do you think?

gautam678 (approved):
Looks good to me. But left minor comments.

docs/CONTRIBUTING.md
dkundel: I'll leave this in until April 30 just because until then we need to support 14 no matter what.

Issue #458, opened 2023-02-01
context.getTwilioClient() error


deshartman:
I have hit an issue with context.getTwilioClient(). I have used it in the past, but now I get the below error from the simple code below. Can someone please help me out here to narrow down the issue. I have included my package.json file for ref.

```
exports.handler = async (context, event, callback) => {

    const restClient = context.getTwilioClient();

    return callback(null, { hello: 'world' });
}

```

This is the error when running this. (I'm using ngrok locally)

```
/Users/dhartman/Development/01_Twilio/twilio-stack/twilio-stack-server/node_modules/@twilio/runtime-handler/dist/dev-runtime/checks/check-auth-token.js:30
        (_a = options.logger) === null || _a === void 0 ? void 0 : _a.error(message, title);
                                                                      ^

TypeError: _a.error is not a function
    at Object.checkForValidAuthToken (/Users/dhartman/Development/01_Twilio/twilio-stack/twilio-stack-server/node_modules/@twilio/runtime-handler/dist/dev-runtime/checks/check-auth-token.js:30:71)
    at Object.getTwilioClient (/Users/dhartman/Development/01_Twilio/twilio-stack/twilio-stack-server/node_modules/@twilio/runtime-handler/dist/dev-runtime/route.js:124:28)
    at exports.handler (/Users/dhartman/Development/01_Twilio/twilio-stack/twilio-stack-server/functions/getJSONData.js:16:32)
    at process.<anonymous> (/Users/dhartman/Development/01_Twilio/twilio-stack/twilio-stack-server/node_modules/@twilio/runtime-handler/dist/dev-runtime/internal/functionRunner.js:74:9)
    at process.emit (node:events:527:28)
    at emit (node:internal/child_process:938:14)
    at processTicksAndRejections (node:internal/process/task_queues:84:21)

```

My package.json file

```{
  "name": "twilio-stack-server",
  "version": "0.0.0",
  "private": true,
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1",
    "start": "twilio-run",
    "deploy": "twilio-run deploy"
  },
  "dependencies": {
    "@twilio/runtime-handler": "1.3.0",
    "twilio": "^3.56"
  },
  "devDependencies": {
    "twilio-run": "^3.5.2"
  },
  "engines": {
    "node": "16"
  }
}```

deshartman:
Here is a simple gitHub example

https://github.com/deshartman/simple-serverless

[...]

PR #459, opened 2023-02-02
Add cross-fork logger for runtime-handler

dkundel:
For every invocation of a local Function we spin up a new forked process. The `logger` instance was passed across the fork but wasn't appropriately serialized/deserialized. As a result you got cryptic error messages like the one in #458 when the process was trying to write to the logger. I changed that by implementing a cross fork logger that has the same interface as the regular instance but then passes the data back via `process.send` the same way that the fork was already communicating with the main thread basically as an RPC to trigger the original logger instance.

**Contributing to Twilio**


packages/runtime-handler/src/dev-runtime/route.ts
```diff
@@ -303,25 +303,46 @@ export function functionPathToRoute(
...
           return;
         }
+
+        if (crossForkLogMessage) {
+          if (
+            config.logger &&
+            typeof config.logger[crossForkLogMessage.level] === 'function'
+          ) {
+            config.logger[crossForkLogMessage.level](
+              // @ts-ignore
```
dkundel, line 328: I'm not too happy about this but typing here would have been quite the pain and in my opinion would distract from the fact that this is just a pass through.

packages/runtime-handler/src/dev-runtime/route.ts
```diff
@@ -303,25 +303,46 @@ export function functionPathToRoute(
...
         debugArgs = [],
+        crossForkLogMessage,
       }: {
         err?: Error | number | string;
         reply?: Reply;
         debugMessage?: string;
         debugArgs?: any[];
+        crossForkLogMessage?: {
+          level: keyof LoggerInstance;
+          args: [string] | [string, number] | [string, string];
```
classactcollin, line 314: I can't imagine these ever changing, but does it make sense to colocation the potential args shapes with the LoggerInstance type? In case the logger instance type is used similarly elsewhere.

packages/runtime-handler/src/dev-runtime/internal/crossForkLogger.ts
```diff
@@ -0,0 +1,35 @@
...
+  log(msg: string, level: number) {
+    this.sendLog('log', msg, level);
+  }
+
+  private sendLog(level: keyof LoggerInstance, ...args: (string | number)[]) {
+    process.send &&
+      process.send({
+        crossForkLogMessage: {
+          level,
+          args: args,
```
classactcollin, line 31:
```suggestion
          args,
```

classactcollin (approved):
Looks good to me! Interesting to read and follow through what's going on here.

packages/runtime-handler/src/dev-runtime/internal/crossForkLogger.ts
```diff
@@ -0,0 +1,35 @@
...
+    this.sendLog('warn', msg, title);
+  }
+
+  error(msg: string, title: string = '') {
+    this.sendLog('error', msg, title);
+  }
+
+  log(msg: string, level: number) {
+    this.sendLog('log', msg, level);
+  }
```
pthirumurthi, line 24: curious, when do we use a [log level](https://github.com/twilio-labs/serverless-toolkit/blob/main/packages/runtime-handler/src/dev-runtime/types.ts#L72)? Also, what does `level` represent here?

packages/runtime-handler/src/dev-runtime/route.ts
```diff
@@ -303,25 +303,46 @@ export function functionPathToRoute(
...
         if (debugMessage) {
           log(debugMessage, ...debugArgs);
           return;
         }
+
+        if (crossForkLogMessage) {
+          if (
+            config.logger &&
+            typeof config.logger[crossForkLogMessage.level] === 'function'
+          ) {
```
gautam678, line 326: nit: `crossForkLogMessage && config.logger && typeof config.logger[crossForkLogMessage.level] === 'function'` 

will typeof config.logger[crossForkLogMessage.level] ever return something other than a function?

gautam678 (approved):

packages/runtime-handler/src/dev-runtime/route.ts
gautam678: args in the sendLog has the type `(string | number)[]` . This was cleaner to read.

pthirumurthi (approved):

vincianechen (approved):

victoray (approved):

Issue #521, opened 2024-05-29
[BUG] [@twilio-labs/serverless-runtime-types@4.0.0] Module '"twilio/lib/rest/Twilio"' has no exported member 'TwilioClientOptions'


ericbn:
**Describe the bug**
Updating @twilio-labs/serverless-runtime-types from version 2.2.3 to 4.0.0 makes tsc fail with `Module '"twilio/lib/rest/Twilio"' has no exported member 'TwilioClientOptions'`. Indeed, I don't see it defined [there](https://github.com/twilio/twilio-node/blob/4.23.0/src/rest/Twilio.ts).

**To Reproduce**
1. Current dependency in package.json is `"@twilio-labs/serverless-runtime-types": "^2.2.3",` (the latest <3 version)
2. tsconfig.json contains `"types": ["./node_modules/@twilio-labs/serverless-runtime-types/index.d.ts"],`
3. Running `npx tsc` works without errors
4. Update to version 4.0.0 with `npm install '@twilio-labs/serverless-runtime-types@4'`
5. Run `npx tsc`
6. See error

**Expected behavior**
tsc fails with
```
node_modules/@twilio-labs/serverless-runtime-types/types.d.ts:5:10 - error TS2305: Module '"twilio/lib/rest/Twilio"' has no exported member 'TwilioClientOptions'.

5 import { TwilioClientOptions } from 'twilio/lib/rest/Twilio';
           ~~~~~~~~~~~~~~~~~~~

node_modules/@twilio-labs/serverless-runtime-types/types.d.ts:415:10 - error TS2305: Module '"twilio/lib/rest/Twilio"' has no exported member 'TwilioClientOptions'.

415 export { TwilioClientOptions } from 'twilio/lib/rest/Twilio';
             ~~~~~~~~~~~~~~~~~~~


Found 2 errors in the same file, starting at: node_modules/@twilio-labs/serverless-runtime-types/types.d.ts:5
```

**Screenshots**

**Desktop**
 - OS: macOS
 - twilio-cli/5.20.0 darwin-x64 node-v18.20.2
 - @twilio-labs/serverless-runtime-types@4.0.0 

**Additional context**
[...]

PR #541, opened 2025-04-18
feat: upgrade twilio v5, use peer deps, enhance tooling

mikefaille:
Upgrades core Twilio dependencies to v5.5.2 across the workspace. Resolves associated type conflicts using peer dependencies. Includes enhancements for create-twilio-function and twilio-run.

Dependency & Type Alignment:
- Upgraded `twilio` dependency to `5.5.2` in `runtime-handler`.
- Changed `twilio` from a direct dependency to a `peerDependency (^5.5.2)` in `serverless-runtime-types` and `twilio-run`.
- This enforces the use of the single root `twilio` instance, resolving type identity conflicts (TS2322/TS2352).
- Updated type definitions within `serverless-runtime-types` (`types.d.ts`) to align with twilio@5.x, primarily using `ClientOpts`.
- Added compatibility export `ClientOpts as TwilioClientOptions`.
- Updated `@twilio/test-dep` alias to `npm:twilio@5.2.2`.
- Added root-level `peerDependency` for `twilio` for workspace checks.

Package Enhancements:
- `create-twilio-function`: Bumped to v3.5.0, updated default dependencies (`twilio`, `typescript`, `serverlessRuntimeTypes`).
- `twilio-run`: fixed `require`->`import` conversion.

Requires a clean install (removing node_modules and lockfile) after merging.

**Contributing to Twilio**


mikefaille:
On my project I have a conflict between  twilio-labs/serverless-runtime-types and twilio:5
This will solve this issue for everyones who want to use twilio functions using node-twilio:5

mikefaille:
That will fix this issue:
https://github.com/twilio-labs/serverless-toolkit/issues/521

package.json
```diff
@@ -26,7 +26,7 @@
     "@changesets/cli": "^2.26.0",
     "@commitlint/cli": "^19.1.0",
     "@commitlint/config-conventional": "^19.1.0",
-    "@twilio/test-dep": "npm:twilio@4.22.0",
+    "@twilio/test-dep": "npm:twilio@5.2.2",
```
jannoteelem: Should it be 5.5.2?

mikefaille: It's the last version. But also the only one I tested with my project. I can't tell if the previous would work.

mikefaille: Oups, you're right. I go update that !

mikefaille: Done

mikefaille:
I forgot; peer dependency doesn't add complexity for the end user. It's actually simpler, with less to worry about. Via the recommended `create-twilio-function`, the user has full control to manage the `twilio` package as they want, without being worried about transitive dependencies.

The next thing to update is the `tsconfig.json` generated by `create-twilio-function` to reflect the latest Node 22 improvements, using `NodeNext` wherever possible (for `module`, `target`, etc.). It's my configuration for Node 18, and it works very well!

mikefaille:
Hi @jannoteelem , 

Circling back on PR #541. I've updated it with the latest from main (commit 6a96529), and the Twilio v5 upgrade / peer dependency changes (to fix #521) are complete.

Recap: This resolves critical type conflicts for Twilio v5 users.

Next Steps:

Could you please re-review the latest changes?
A workflow also needs maintainer approval.
Happy to address any feedback. Thanks!

yinzara:
Would love to see this updated to latest deps and conflicts resolved so it can be reviewed and merged. It fixes multiple issues including #521

mikefaille:
@yinzara I can update it if you want.

yinzara:
I would love that though I can't promise anyone at Twilio is actually going to do anything about it seeing as they've totally ignored it before.

mikefaille:
It's my only worry to use Twilio API... Or putting time to maintain something that would be ingnored. But I'm glad you like.

packages/create-twilio-function/package.json
```diff
@@ -24,7 +24,7 @@
   },
   "license": "MIT",
   "devDependencies": {
-    "@twilio/runtime-handler": "^2.0.3",
+    "@twilio/runtime-handler": "^2.1.0",
```
yinzara, line 27: Unfortunately this will break deployment period because of #564

packages/twilio-run/package.json
yinzara, line 1: Mind adding the "24.x" engine to the list of engines in this file as well so we can fix #563

package.json
```diff
@@ -45,6 +45,9 @@
     "typedoc": "^0.28.14",
     "typescript": "^5.3.3"
   },
+  "peerDependencies": {
+			"twilio": "^5.6.0"
```
yinzara: "^5.6.0" is going to cause using "6.1.0" (the current latest version ) to fail the peerDependencies check.  Maybe we should use ">=5.6.0"? Thoughts from the Twilio internal team?

packages/create-twilio-function/src/create-twilio-function/versions.js
```diff
@@ -1,13 +1,13 @@
...
-  twilio: '5.0.3',
+  twilio: '5.6.0',
   twilioRuntimeHandler: pkgJson.devDependencies[
     '@twilio/runtime-handler'
   ].replace(/[\^~]/, ''),
   twilioRun: pkgJson.dependencies['twilio-run'],
   node: '^20.x || ^22.x',
-  typescript: '^5.3.3',
-  serverlessRuntimeTypes: '^4.0.0',
+  typescript: '^5.8.0',
```
yinzara: I'm curious why you choose 5.8.0 when 6.0.3 is the latest version. Was there a dependency somewhere else that failed with 6.0.3?

